PR #27, opened 2021-11-17
Fixes 26 - Adds config file creation logic in case it doesn't exist

harishasan:


src/client-utils/auth-manager.ts
```diff
@@ -26,5 +26,11 @@ export async function getAuthInfo(configDir: string): Promise<AuthInfo | null> {
...
  */
 export async function setAuthInfo(credentials: AuthInfo, configDir: string): Promise<void> {
-  return await fs.writeFile(path.join(configDir, "config.json"), JSON.stringify(credentials));
+  
+  const configFilePath = path.join(configDir, "config.json");
+
+  if(!fs.existsSync(configFilePath))
+    fs.createFileSync(configFilePath)
+
+  return await fs.writeFile(configFilePath, JSON.stringify(credentials));
```
safeerraees, line 35: Instead of checking file exists we can use the "w" flag option of the writeFile function to create the file if it doesn't exist. Although, it is default in Nodejs documentation, we might need to investigate why the default flag is not working
Reference:  
https://nodejs.org/api/fs.html#fs_file_system_flags

image
image

safeerraees (changes requested):

src/client-utils/auth-manager.ts
harishasan: @safeerraees this is the reason it is failing https://stackoverflow.com/a/50927704

So we can also do 
```
await mkdirp(configDir);
return await fs.writeFile(path.join(configDir, "config.json"), JSON.stringify(credentials), { flag: "w" });
```

Your call however you want to do this.

safeerraees (approved):

PR #30, opened 2023-09-21
Migrate CLI to use base url of subdomain api.apimatic.io

piyush429:
Migrate the CLI to use the new base url of api calls from "www.apimatic.io/api" to "api.apimatic.io"


Muhammad-Sajjad (approved):
Looks good to me.

thehappybug (approved):

Issue #10, opened 2021-10-11
[Meta] CLI V3


thehappybug:
## Tier 1

These commands should go out in the first release. For more details about these commands, check out the [spec document](https://docs.google.com/document/d/1QB6EL62giMJtqVNmHtW7mkjPVAzMKU6QZTGU9fimNjU/edit?usp=sharing).

Priority goes to implementing the standalone commands first that is the commands that don't deal with API Entity/Group IDs. Our goal is to first cover the cases where the CI/CD pipeline creates artifacts without storing state in the website/dashboard.

Here's an priority-sorted list of commands (only implement the variant of these commands that does not deal with APIs on the website):

* [x] `auth:login`
* [x] `auth:logout`
* [x] `auth:status`
* [x] `api:transform`
* [x] `api:validate`
* [x] `sdk:generate`
* [x] `portal:generate`

After these commands are done, we'll tackle the state-oriented commands that is the ones where the API spec is stored in the dashboard. 

* [x] `api:import`
* [x] `portal:publish`

## Tier 2

And these which have lower priority than the ones above. Also, they have a lot more unknown (issues to be resolved) compared to the commands above:

* [ ] `api:list`
* [x] `api:set`
* [ ] `api:transform`*
* [x] `api:validate`*
* [ ] `sdk:list`
* [x] `sdk:generate`*
* [x] `sdk:download`
* [ ] `package:publish`
* [ ] `package:list`
* [ ] `package:info`
* [ ] `package:update`
* [x] `portal:unpublish`
* [x] `portal:scaffold`
* [x] `portal:serve`
* [ ] `issue:report`

_(*) This command also has a stateless version._

## Tier 3
- [x] portal:quickstart
- [ ] sdk:quickstart
- [x] portal:add --toc 
  - Help the user generate an exhaustive list of all endpoints in the spec 
- [ ] portal:add --recipe
- [ ] portal:add --guide

safeerraees:
@shayanjalil We need APIs for api:list,, api:transform with api entity id for tier 2 development.

safeerraees:
In portal unpublish API if an API Entity Id is provided against which there is no portal published then it should return 404, but currently, it returns 204.

sohaibtariq:
We are skipping package publishing for now since we plan to upgrade this entire module in the apimatic app

sohaibtariq:
Updated scope of tier 2 release:
- portal:scaffold
- portal:serve
- improved logging of responses
- usage analytics

[...]

PR #33, opened 2025-02-03
feat(portal): adds portal quickstart and serve commands

aliasghar98:
Implements a part of #10 .

## What was Added?
- Added the portal:quickstart command for helping new users get started with APIMatic Docs-as-Code solution.
- Added the portal:serve command to allow serving of static portals generated using Docs-as-Code locally.


src/commands/portal/serve.ts
```diff
@@ -0,0 +1,134 @@
...
+  static examples = [
+    '$ apimatic portal:serve --source="./" --destination="./generated_portal" --port=3000 --open --reload'
+  ];
+
+  async run() {
+    const { flags } = this.parse(PortalServe);
+    const ignoredPaths = flags.ignore.split(",").map((path) => path.trim());
+    const portalDir = flags.destination;
+    const sourceDir = flags.source;
+    const port = flags.port;
```
beenishkhan603, line 68: Please add a check on port to prevent invalid values from being used. (isNaN(port) || port <= 0 || port >= 65536)  something like this

src/commands/portal/quickstart.ts
```diff
@@ -0,0 +1,138 @@
...
+    controller: PortalQuickstartController,
+    specFile: SpecFile,
+    apiValidationSummary: ApiValidationSummary,
+    languages: string[]
+  ): Promise<string> {
+    const directory = await prompts.buildDirectoryPrompt();
+
+    prompts.displayBuildDirectoryGenerationMessage();
+
+    await controller.setupBuildDirectory(directory, specFile, apiValidationSummary, languages);
```
beenishkhan603: add such function calls in try catch to prevent the code from crashing

src/controllers/api/validate.ts
```diff
@@ -1,25 +1,40 @@
...
 
-  cli.action.start("Validating specification file");
   if (file) {
-    const fileDescriptor = new FileWrapper(fs.createReadStream(file));
-    validation = await apiValidationController.validateAPIViaFile(ContentType.EnumMultipartformdata, fileDescriptor);
+    const fileStatus = fs.statSync(file);
+    if (fileStatus.isDirectory()){
+      const tempDir = await createTempDirectory();
+
+      try {
```
beenishkhan603, line 18: Please add catch block

src/commands/portal/serve.ts
```diff
@@ -0,0 +1,134 @@
...
+          process.exit(0);
+        });
+      };
+
+      process.on("SIGINT", shutdown);
+      process.on("SIGTERM", shutdown);
+    });
+  }
+
+  private handleError(error: any) {
```
beenishkhan603: try avoid using any in the code

PR #57, opened 2025-04-09
fix(portal): various improvements for portal commands

### LINKED ISSUE #55: Portal quickstart terminates on invalid input
If there is an error reading the provided API definition file, the quickstart flow terminates instead of asking the user to provide a valid Api Definition
Image

### LINKED ISSUE #56: Improvements to Portal Commands
## Quickstart

- [x] [Urgent] Placeholder text for spec import prompt becomes actual text upon pressing 'Enter' to use sample API. This is an issue of the prompts package that we are using i.e. `@clack/prompts`. One of their recent releases introduced this bug and we are waiting for a resolution from their end.
- [x] Use the same default port as in the serve command. Using two different default ports is inconsistent behaviour.
- [x] Create the portal build files in current working directory instead of a separate `apimatic-quickstart-portal` folder.
- [x] Trim user inputs for each prompt.

## Generate

- [x] Temporary file created containing the portal build files does not ignore already present portal artifacts in the directory.
- [x] Generated artifacts folder needs to have consistent naming. Currently we are using `api-portal` and `generated_portal`.
- [x] Prompt the user for overwriting existing portal artifacts folder in case it's already present.
- [x] Improve error handling.

## Serve

- [x] Running multiple servers incorrectly throws exception about port already in use.

aliasghar98:
Closes #55 , Closes #56 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,4 @@ $RECYCLE.BIN/
 
 lib/
 tmp/
+.vscode/launch.json
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ USAGE
   $ apimatic portal:serve [-p <value>] [-d <value>] [-s <value>] [-o] [--no-reload] [-i <value>] [--auth-key <value>]
 
 FLAGS
-  -d, --destination=<value>  [default: ./api-portal] Directory to store and serve the generated portal.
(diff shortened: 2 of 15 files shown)


aliasghar98:
Replaced by https://github.com/apimatic/apimatic-cli/pull/59.

PR #129, opened 2025-07-30
fix(apimatic-cli): Resolved issue where paths with quotes caused error

sohail2721:


src/prompts/portal/recipe/new-recipe.ts
```diff
@@ -38,20 +38,32 @@ export class PortalRecipePrompts {
...
   public async buildConfigFilePathPrompt(buildDirectoryPath: string): Promise<string> {
     const buildConfigFilePath = await text({
       message: `⚠️  APIMATIC-BUILD.json is required and was not found in "${buildDirectoryPath}".\nPlease enter the path to your build config file (relative to this directory):`,
       validate: (filePath) => {
         if (!filePath) {
           return "Build config file path cannot be empty. Please provide a valid file path.";
         }
-
-        if (!filePath.endsWith(".json")) {
+        const cleanedPath = this.removeQuotes(filePath.trim() ?? "");
```
saeedjamshaid: is there a chance that `filePath` might be null or undefined here? If that's the case then it is not safe to call trim() on it.

saeedjamshaid (changes requested):

saeedjamshaid (approved):

PR #140, opened 2025-08-04
fix: remove ignore flag from portal serve

aliasghar98:
What was added?
- Added check for existing portal to portal:serve.

What was changed?
- Removed ignore flag from portal:serve.


src/actions/portal/serve.ts
```diff
@@ -13,6 +13,7 @@ export class PortalServeAction {
   protected readonly docsPortalService: PortalService;
   private readonly GENERATED_PORTAL_ARTIFACTS_ZIP_FILENAME = ".generated_portal.zip";
   private readonly GENERATED_BUILD_INPUT_ZIP_FILENAME = ".portal_source.zip";
+  private readonly GENERATED_SPEC_SOURCE_ZIP_FILENAME = ".spec_source.zip";
```
Shield-Jaguar: remove these files as they are no longer needs to be ignored

Shield-Jaguar (changes requested):

src/infrastructure/services/portal-service.ts
```diff
@@ -77,8 +77,8 @@ export class PortalService {
     }
   }
 
-  public async generateSdl(specPath: string, configDir: string): Promise<Result<Sdl, string>> {
-    if (!(await fs.pathExists(specPath))) {
+  public async generateSdl(specPath: FilePath, configDir: string): Promise<Result<Sdl, string>> {
+    if (!(await fs.pathExists(specPath.toString()))) {
```
Shield-Jaguar: either use file service or remove this check.

Shield-Jaguar (changes requested):

src/infrastructure/services/portal-service.ts
```diff
@@ -77,18 +77,15 @@ export class PortalService {
...
-      return Result.failure("Spec file doesn't exist");
-    }
-
+  public async generateSdl(specPath: FilePath, configDir: string): Promise<Result<Sdl, string>> {
     const authInfo: AuthInfo | null = await getAuthInfo(configDir);
     const authorizationHeader = this.createAuthorizationHeader(authInfo, null);
     const client = this.createApiClient(authorizationHeader);
     const transformationController = new TransformationController(client);
 
     try {
```
Shield-Jaguar, line 89: The stream here is not closed, 

get stream function should not exist in api-service classes. instead pass stream directly to this method.

Shield-Jaguar (changes requested):

Shield-Jaguar (approved):

PR #143, opened 2025-08-06
feat: Added responses for generate-via-file

sohail2721:


src/infrastructure/services/portal-service.ts
```diff
@@ -157,31 +160,42 @@ export class PortalService {
     }
   };
 
-  private parseErrorResponse = async (error: unknown): Promise<ErrorResponse> => {
-    if (error instanceof Error && "body" in error) {
-      const stream = (error as { body: NodeJS.ReadableStream }).body;
-      return await parseStreamBodyToJson(stream);
+  private parseErrorResponse(errorMessage: string | undefined): string {
```
saeedjamshaid: rename this method to parseBadRequestResponse

saeedjamshaid (changes requested):

saeedjamshaid (approved):

PR #167, opened 2025-08-18
fix: issue with portal generation error parsing in case of 400

saeedjamshaid:
closes https://github.com/apimatic/apimatic-io/issues/1805


aliasghar98 (approved):
LGTM

PR #180, opened 2025-08-19
refactor: improved portal quickstart to use command-action structure

aliasghar98:


.gitignore
```diff
@@ -180,3 +180,4 @@ test-source/
 test-destination/
 test-portal/
 /.idea/
+guidelines
```
Shield-Jaguar: please what is the standard or de-facto standard for strong copilot config

src/utils/utils.ts
```diff
@@ -12,68 +14,6 @@ export const createTempDirectory = async () => {
...
-      directoryStructure[
-        descriptions[path.normalize(relativePath)] ? `${item} : ${descriptions[path.normalize(relativePath)]}` : item
-      ] = null;
-    }
-  });
-
-  return directoryStructure;
-};
-
 export const isValidUrl = (input: string): boolean => {
```
Shield-Jaguar: either call the urlpath method here or use the logic defined here. there should be no duplication

src/types/resource-context.ts
```diff
@@ -0,0 +1,55 @@
...
+          "Unable to download the API Definition. Please verify that the provided URL is correct and publicly accessible. "
+        );
+      }
+
+      const destinationFilePath = new FilePath(this.tempDirectory, urlPath.fileName());
+      await this.fileService.writeFile(destinationFilePath, downloadFileResult.value);
+      return ok(destinationFilePath);
+    }
+
+    //TODO: Create factory method for preparing FilePath from a string.
```
Shield-Jaguar: always add a space before comments 
// TODO: etc

src/types/portal/quickstart.ts
```diff
@@ -1,19 +1,3 @@
-export type LoginCredentials = {
-    email: string;
-    password: string;
+export interface DirectoryNode {
+  [key: string]: DirectoryNode | string | null | undefined;
```
Shield-Jaguar, line 2: are we still using that

src/infrastructure/file-service.ts
```diff
@@ -1,12 +1,18 @@
...
 import fsExtra from "fs-extra";
+import filetype from "file-type";
 import * as path from "path";
 import { FilePath } from "../types/file/filePath.js";
 import { DirectoryPath } from "../types/file/directoryPath.js";
 import { pipeline } from "stream";
 import { promisify } from "util";
 
 export class FileService {
+  public async isZipFile(file: FilePath): Promise<boolean> {
```
Shield-Jaguar: what is this method doing? just check if it has a zip extension? then it should exit here.

src/infrastructure/file-service.ts
```diff
@@ -16,6 +22,10 @@ export class FileService {
     }
   }
 
+  public async ensureDirectoryExists(dir: DirectoryPath): Promise<void> {
```
Shield-Jaguar: This type of method creates perception that you don't know what you are doing. Try to remove this method.

src/infrastructure/file-service.ts
```diff
@@ -61,6 +71,13 @@ export class FileService {
     }
   }
 
+  public async deleteDirectory(dirPath: DirectoryPath): Promise<void> {
```
Shield-Jaguar, line 64: we already have a method called clean directory. try to use that if possible

src/infrastructure/services/telemetry-service.ts
```diff
@@ -19,8 +19,7 @@ type TelemetryPayload = {
 export class TelemetryService {
   private readonly apiService = new ApiService();
 
-  constructor(private readonly configDirectory: string) {
-  }
+  constructor(private readonly configDirectory: string) {}
```
Shield-Jaguar: THis should be directoryPath

src/infrastructure/services/file-download-service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { UrlPath } from "../../types/file/urlPath.js";
+import { err, ok, Result } from "neverthrow";
+
+export class FileDownloadService {
+  public async downloadFile(url: UrlPath): Promise<Result<NodeJS.ReadableStream, string>> {
+    try {
+      const response = await axios.get(url.toString(), { responseType: "stream" });
+      return ok(response.data as NodeJS.ReadableStream);
+    } catch {
+      return err(
```
Shield-Jaguar: it should return serviceError

Shield-Jaguar (changes requested):

src/types/portal/quickstart.ts
aliasghar98: Yes.

.gitignore
aliasghar98: Removed this for now. Will update it later.

src/infrastructure/file-service.ts
aliasghar98: Moved to `fileName.ts`

src/infrastructure/file-service.ts
aliasghar98: Removed this method.

src/infrastructure/file-service.ts
aliasghar98: Deleted this method and used that.

src/infrastructure/services/file-download-service.ts
aliasghar98: Done.

src/infrastructure/services/telemetry-service.ts
aliasghar98: Done.

src/types/resource-context.ts
aliasghar98: Done.

Shield-Jaguar (approved):

PR #209, opened 2025-09-18
feat: updated messaging and made spec flag optional in sdk quickstart

aliasghar98:


src/actions/sdk/quickstart.ts
```diff
@@ -169,8 +169,7 @@ export class SdkQuickstartAction {
         this.prompts.sdkOpenedInEditor();
       }
 
-      this.prompts.nextSteps(specDirectory, language);
-
+      this.prompts.nextSteps(language, inputDirectory.toString() !== process.cwd() ? specDirectory : undefined);
```
Shield-Jaguar: we cannot use toString() on application layer. 

process.cwd() is also infra code.

src/prompts/sdk/quickstart.ts
```diff
@@ -179,17 +180,22 @@ export class SdkQuickstartPrompts {
...
   public sdkOpenedInEditor() {
-    log.info("Opened the SDK directory in VS Code.");
+    log.info(
+      "Opened the SDK directory in VS Code. To get started with your SDK, review the README file."
+    );
   }
 
-  public nextSteps(specDirectory: DirectoryPath, language: string): void {
-    const message = `- Run the command '${f.cmdAlt(
+  public nextSteps(language: string, specDirectory?: DirectoryPath): void {
```
Shield-Jaguar: language should be enum or primitive class

Shield-Jaguar (changes requested):

Shield-Jaguar (approved):

PR #217, opened 2025-10-14
feat: add support for pruning spec and use v2 validation endpoint

sohail2721:


Shield-Jaguar (approved):
The code needs refactoring.

PR #221, opened 2025-11-10
fix: avoid adding empty events or models section in toc

Ayeshas09:
This PR fixes an issue where the TOC was adding Events and Models sections even when they were not present in the spec file. Now, these sections are generated only if the spec file contains the relevant data.

### DIFF
--- a/src/application/portal/toc/toc-structure-generator.ts
+++ b/src/application/portal/toc/toc-structure-generator.ts
@@ -128,7 +128,7 @@ export class TocStructureGenerator {
     }
     return [
       {
-        generate: null,
+        generate: "Callbacks",
         from: 'callbacks'
       }
     ];
@@ -159,7 +159,7 @@ export class TocStructureGenerator {
     }
     return [
       {
-        generate: null,
+        generate: "Webhooks",
         from: 'webhooks'
       }
     ];
--- a/src/application/portal/toc/toc-structure-generator.ts
+++ b/src/application/portal/toc/toc-structure-generator.ts
@@ -12,14 +12,36 @@ import {
 
 // TODO: Refactor
 
+type Endpoints = {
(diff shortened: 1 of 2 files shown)


src/application/portal/toc/toc-structure-generator.ts
```diff
@@ -77,61 +103,86 @@ export class TocStructureGenerator {
...
-    return {
-      group: 'Callbacks',
-      items: Array.from(callbackGroups).map(([groupName, eventList]) => ({
-        group: groupName,
-        items: eventList
-      }))
-    };
+    return [
+      {
+        generate: null,
```
aliasghar98: Shouldn't it be `generate: Callbacks` instead of `generate: null`? Please check the scenario where a spec containing WH/CB is given but the flags i.e. `expand-webhooks` and `expand-callbacks` are not set.

src/application/portal/toc/toc-structure-generator.ts
```diff
@@ -77,61 +103,86 @@ export class TocStructureGenerator {
...
-    return {
-      group: 'Webhooks',
-      items: Array.from(webhookGroups).map(([groupName, eventList]) => ({
-        group: groupName,
-        items: eventList
-      }))
-    };
+    return [
+      {
+        generate: null,
```
aliasghar98: Please see my comment about Callbacks where I mentioned `generate: null` added instead of `generate: Callbacks`. I believe this should also be `generate: Webhooks` instead of `generate: null`, no?

aliasghar98 (changes requested):

src/application/portal/toc/toc-structure-generator.ts
Ayeshas09: Yes, I have updated it

aliasghar98 (approved):
LGTM

PR #224, opened 2025-11-10
fix: failing pruning api call

aliasghar98:
What was added?

- Added a fix for the failing pruning API call to set the endpoint URL correctly.


saeedjamshaid (approved):
LGTM

PR #227, opened 2025-11-25
fix: broken quickstart sample flow and empty events in toc

aliasghar98:
What was Added?

- Added a fix for only adding the events section in TOC when an API spec containing Webhooks/Callbacks was provided.
- Added a missing check for verifying validation was successful when checking unallowed features.


saeedjamshaid (approved):
LGTM

PR #232, opened 2026-01-02
feat: update status response handling for portal generation

aliasghar98:
What was added?

- Updated portal generation checks for new status response body.


saeedjamshaid (approved):
LGTM

PR #245, opened 2026-03-06
fix: update sdk quickstart to create minimal build

### LINKED ISSUE #246: SDK Quickstart Generates Portal-Style Build Instead of Minimal Spec-Only Build
Currently, running sdk quickstart generates a build structure similar to portal quickstart. This build includes additional files and directories that are required for portal generate, but are unnecessary for SDK generation.

### Expected Behavior
When running sdk quickstart, the generated build should contain only the minimal structure necessary for SDK generation, specifically the spec directory.

### Current Behavior
sdk quickstart produces the same build structure as portal quickstart, which includes additional content intended for portal generation.

### Suggested Fix
Update sdk quickstart to generate a minimal build containing only the spec directory required for SDK generation, rather than replicating the full portal quickstart structure.

Ayeshas09:
Closes #246 

What was added?

- Updated sdk quickstart to generate a minimal build containing only the spec directory required for SDK generation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@apimatic/cli",
-    "version": "1.1.0-beta.6",
+    "version": "1.1.0-beta.7",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@apimatic/cli",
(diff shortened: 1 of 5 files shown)


aliasghar98 (approved):
LGTM

PR #248, opened 2026-03-10
feat: add api-version flag in sdk generate

### LINKED ISSUE #249: Support API version selection in sdk generate for multi-versioned builds
Currently, when the `sdk generate` command is executed on a versioned build, the SDK is always generated using the first available API version.

### Expected Behavior
Users should be able to specify the API version when running the sdk generate command.

### Proposed Solution
Introduce a new flag: `--api-version`
This flag would allow users to specify the desired API version when generating the SDK.

Ayeshas09:
Closes #249 

What was added?

- Added optional `api-version` flag to `sdk generate` to allow selecting the API version for multi-versioned builds.

### DIFF
--- a/src/actions/sdk/generate.ts
+++ b/src/actions/sdk/generate.ts
@@ -37,26 +37,30 @@ export class GenerateAction {
     }
 
     const versionedBuildContext = new VersionedBuildContext(buildDirectory);
-    if (await versionedBuildContext.isVersioned()) {
-      const versionDirs = await versionedBuildContext.getVersionDirectories();
-      if (versionDirs.length === 0) {
-        this.prompts.versionedBuildEmpty(await versionedBuildContext.getVersionsDirectory());
+    const versionedBuildResult = await versionedBuildContext.validate();
+    if (versionedBuildResult.isValid) {
+      if (versionedBuildResult.versions.length === 0) {
+        this.prompts.versionedBuildEmpty(versionedBuildResult.versionsDirectory);
         return ActionResult.failed();
       }
 
-      const resolvedDirectory = versionDirs.length === 1
-        ? versionDirs[0]
-        : apiVersion
-          ? await versionedBuildContext.resolveVersionDirectory(apiVersion)
-          : await this.prompts.selectVersion(versionDirs);
+      if (apiVersion && !versionedBuildResult.versions.includes(apiVersion)) {
+        this.prompts.versionNotFound();
+          return ActionResult.failed();
+        }
 
-      if (!resolvedDirectory) {
-        if (apiVersion) this.prompts.versionNotFound();
-        return apiVersion ? ActionResult.failed() : ActionResult.cancelled();
+      const version = versionedBuildResult.versions.length === 1
+        ? versionedBuildResult.versions[0]
+        : apiVersion
+          ? apiVersion
+          : await this.prompts.selectVersion(versionedBuildResult.versions);
+
+      if (!version) {
+        return ActionResult.cancelled();
       }
 
-      buildDirectory = resolvedDirectory;
-      sdkDirectory = sdkDirectory.join(resolvedDirectory.leafName());
+      buildDirectory = versionedBuildResult.versionsDirectory.join(version);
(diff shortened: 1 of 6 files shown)


src/commands/sdk/generate.ts
```diff
@@ -26,6 +26,9 @@ Supports multiple programming languages including Java, C#, Python, JavaScript,
       char: "d",
       description: "Directory where the SDK will be generated"
     }),
+    "api-version": Flags.string({
+      description: "Version of the API documentation to use for SDK generation (if multiple versions exist)"
```
aliasghar98: Maybe we can update this to "Version of the API to use for SDK generation (if multiple versions exist)"

src/types/versioned-build-context.ts
```diff
@@ -1,27 +1,52 @@
...
+  private buildConfig: Record<string, unknown> | undefined;
 
   constructor(buildDirectory: DirectoryPath) {
-    this.versionedDocsDir = buildDirectory.join("versioned_docs");
+    this.buildDirectory = buildDirectory;
   }
 
-  public async exists(): Promise<boolean> {
-    return this.fileService.directoryExists(this.versionedDocsDir);
+  private async getBuildConfig(): Promise<Record<string, unknown> | undefined> {
```
aliasghar98: Please refactor this method by using BuildContext class as a reference.

aliasghar98 (approved):

src/actions/sdk/generate.ts
```diff
@@ -28,22 +28,35 @@ export class GenerateAction {
...
-      if (!resolvedDirectory) {
-        this.prompts.versionedBuildEmpty();
+    if (await versionedBuildContext.isVersioned()) {
+      const versionDirs = await versionedBuildContext.getVersionDirectories();
+      if (versionDirs.length === 0) {
+        this.prompts.versionedBuildEmpty(await versionedBuildContext.getVersionsDirectory());
         return ActionResult.failed();
       }
+
+      const resolvedDirectory = versionDirs.length === 1
```
asadali214: @Ayeshas09 Please refactor this ternary operation to reduce its complexity. Also, if `versionDirs.length > 0` then priority should only be given to `apiVersion` if provided

asadali214: Currently, if there is only one version directory and the user provided an invalid version in apiVersion flag then you are not failing with versionNotFound error instead you are rejecting the user provided version and going with the one found

src/types/versioned-build-context.ts
```diff
@@ -1,27 +1,52 @@
...
+    const versionsDir = await this.getVersionsDirectory();
+    if (!(await this.fileService.directoryExists(versionsDir))) {
+      return [];
+    }
+    return await this.fileService.getSubDirectoriesPaths(versionsDir);
+  }
+
+  public async resolveVersionDirectory(apiVersion: string): Promise<DirectoryPath | null> {
+    const versionDirs = await this.getVersionDirectories();
+    return versionDirs.find((dir) => dir.leafName() === apiVersion) ?? null;
```
asadali214: nits: use undefined instead of null

src/actions/sdk/generate.ts
```diff
@@ -28,22 +28,35 @@ export class GenerateAction {
...
       this.prompts.sameBuildAndSdkDir(buildDirectory);
       return ActionResult.failed();
     }
 
     const versionedBuildContext = new VersionedBuildContext(buildDirectory);
-    if (await versionedBuildContext.exists()) {
-      const resolvedDirectory = await versionedBuildContext.getResolvedBuildDirectory();
-      if (!resolvedDirectory) {
-        this.prompts.versionedBuildEmpty();
+    if (await versionedBuildContext.isVersioned()) {
```
asadali214: The versioned buildContext should only expose a validate function, which will be responsible for returning a versionedBuildValidationResult.

This will avoid the repeated dependency on a versionedBuildContext instance (i.e. being used 4 times here)

asadali214 (changes requested):

asadali214 (approved):

PR #251, opened 2026-03-12
feat: add support for saving and applying sdk customizations

Ayeshas09:


src/actions/portal/serve.ts
```diff
@@ -38,7 +38,7 @@ export class PortalServeAction {
     onAfterServe?: () => void
   ): Promise<ActionResult> {
     const generatePortalAction = new GenerateAction(this.configDir, this.commandMetadata, this.authKey);
-    const result = await generatePortalAction.execute(buildDirectory, portalDirectory, true, false);
+    const result = await generatePortalAction.execute(buildDirectory, portalDirectory, true, false, false);
```
asadali214: Why is displaying messages turned off?

.vscode/launch.json
```diff
@@ -19,7 +19,10 @@
             // "auth:login",
             // "auth:status"
             // "portal:quickstart",
-            // "portal:generate",
+            "portal:generate",
+            "--input=D:/calculator",
+            "--destination=D:/calculator-portal",
```
asadali214: Remove the changes from the launch.json file

src/actions/sdk/generate.ts
```diff
@@ -8,11 +8,16 @@ import { SdkGeneratePrompts } from "../../prompts/sdk/generate.js";
...
+import { FilePath } from "../../types/file/filePath.js";
+import { MergeSourceTreeAction } from "./merge-source-tree.js";
+import { BuildContext } from "../../types/build-context.js";
+import { FileService } from "../../infrastructure/file-service.js";
 
 export class GenerateAction {
   private readonly prompts: SdkGeneratePrompts = new SdkGeneratePrompts();
   private readonly portalService: PortalService = new PortalService();
+  private readonly fileService: FileService = new FileService();
+  private readonly mergeSourceTreeAction: MergeSourceTreeAction = new MergeSourceTreeAction();
```
asadali214: Since merge source tree action, is only going to be a part of sdk generate action, there is no need for a separate MergeSourceTreeAction class

src/actions/sdk/generate.ts
```diff
@@ -82,22 +69,41 @@ export class GenerateAction {
       const tempContext = new TempContext(tempDirectory);
       const buildZipPath = await tempContext.zip(buildDirectory);
 
+      // TODO: pass build file
```
asadali214: nits: remove TODO

src/actions/sdk/generate.ts
```diff
@@ -82,22 +69,41 @@ export class GenerateAction {
...
       const response = await this.prompts.generateSDK(
         this.portalService.generateSdk(buildZipPath, language, this.configDir, this.commandMetadata, this.authKey)
       );
 
-      // TODO: this should be service error
+      // TODO: this should be a service error
```
asadali214: Check if this TODO can also be removed since the `response.error` is always a service error in this case

src/actions/sdk/save-changes.ts
```diff
@@ -0,0 +1,126 @@
...
+  private readonly gitService = new GitService();
+
+  constructor(
+    private readonly configDir: DirectoryPath,
+    private readonly commandMetadata: CommandMetadata
+  ) {}
+
+  public async execute(buildDirectory: DirectoryPath, updatedSdkDirectory: DirectoryPath, language: Language): Promise<ActionResult> {
+    const buildContext = new BuildContext(buildDirectory);
+        if (!(await buildContext.validate())) {
```
asadali214: nits: indentation issues

src/actions/sdk/save-changes.ts
```diff
@@ -0,0 +1,126 @@
...
+          this.prompts.srcDirectoryEmpty(buildDirectory);
+          return ActionResult.failed();
+        }
+
+    if (!(await this.fileService.directoryExists(updatedSdkDirectory))) {
```
asadali214: Add validation for versionedBuildContext as well

src/actions/sdk/save-changes.ts
```diff
@@ -0,0 +1,126 @@
...
+        this.prompts.operationCancelled();
+        return ActionResult.cancelled();
+      }
+
+      await this.fileService.copyDirectoryExcluding(updatedSdkDirectory, tempDirectory, [".git"]);
+
+      const latestStatuses = await this.gitService.getModifiedFilesWithStatus(tempDirStr);
+      const allChangedFiles = latestStatuses.map(fs => fs.file);
+      await this.gitService.stageFiles(tempDirStr, allChangedFiles);
+      await this.gitService.commit(tempDirStr, "add customizations");
```
asadali214: The commit messages should always follow commit linting standards

src/actions/sdk/save-changes.ts
```diff
@@ -0,0 +1,126 @@
...
+      await this.gitService.checkoutToMain(tempDirStr);
+      await this.zipService.archive(
+        new DirectoryPath(path.join(tempDirStr, ".git")),
```
asadali214: Please check, at this moment, whether the .git directory still contains the review dir?

src/commands/sdk/generate.ts
```diff
@@ -26,15 +26,20 @@ Supports multiple programming languages including Java, C#, Python, JavaScript,
       char: "d",
       description: "Directory where the SDK will be generated"
     }),
-    "api-version": Flags.string({
-      description: "Version of the API to use for SDK generation (if multiple versions exist)"
+    "skip-apply-source-tree": Flags.boolean({
```
asadali214: Change name to `skip-changes`

src/commands/sdk/generate.ts
```diff
@@ -26,15 +26,20 @@ Supports multiple programming languages including Java, C#, Python, JavaScript,
...
+      description: "Do not apply the source tree customizations to the generated SDK"
     }),
     ...FlagsProvider.force,
     zip: Flags.boolean({
       default: false,
       description: "Download the generated SDK as a .zip archive"
     }),
-    ...FlagsProvider.authKey
+    ...FlagsProvider.authKey,
+    "build-source-tree": Flags.boolean({
```
asadali214: Change name to `track-changes`

src/infrastructure/launcher-service.ts
```diff
@@ -8,10 +8,32 @@ import isInCi from "is-in-ci";
 import { DirectoryPath } from "../types/file/directoryPath.js";
 
 export class LauncherService {
-  public async openFolderInIde(directoryPath: DirectoryPath, fileToOpen: FilePath): Promise<boolean> {
+  public async openFolderInIde(directoryPath: DirectoryPath, filesToOpen: FilePath | FilePath[]): Promise<boolean> {
```
asadali214: refactor: use ...args pattern

asadali214 (changes requested):

src/infrastructure/tmp-extensions.ts
```diff
@@ -1,8 +1,17 @@
...
 ): Promise<T> {
-  return withDir(results => fn(new DirectoryPath(results.path)), { unsafeCleanup: true });
+  const tmpDir = await dir({ unsafeCleanup: true });
+  try {
+    return await fn(new DirectoryPath(tmpDir.path));
+  } finally {
+    try {
+      await tmpDir.cleanup();
+    } catch {
+      // Ignore cleanup errors to avoid failing when the temp directory can't be deleted
```
Shield-Jaguar: why this cannot be deleted. The Old code is fully tested and in production? are we not properly disposing the file pointers? This seems like a hack to me.

Shield-Jaguar (changes requested):

src/hooks/not-found.ts
```diff
@@ -46,7 +46,7 @@ const hook: Hook.CommandNotFound = async function (opts) {
 
   if (response) {
     const confirmedSuggestion = suggestion!;
-    let argv = opts.argv?.length ? opts.argv : opts.id.split(":").slice(confirmedSuggestion.split(":").length);
+    let argv = opts.argv ?? [];
```
Shield-Jaguar, line 49: check with Sohail if this change is valid.

src/types/build-context.ts
```diff
@@ -40,5 +40,17 @@ export class BuildContext {
   public async deleteWorkflowDir() {
     await this.fileService.deleteDirectory(this.buildDirectory.join(".github"));
   }
+
+  public getSdkSourceTreePath(language: string): FilePath | undefined {
```
Shield-Jaguar: This function should not return Undefined.

src/types/build-context.ts
```diff
@@ -40,5 +40,17 @@ export class BuildContext {
...
+    return FilePath.create(this.buildDirectory.join("sdk-source-tree", `.${language}`).toString());
+  }
+
+  public async hasSdkSourceTree(language: string): Promise<boolean> {
```
Shield-Jaguar, line 49: THis is confusing. the combination of getSdkSourceTreePath and hasSdkSourceTree.

src/types/events/sdk-conflicts-resolved.ts
```diff
@@ -0,0 +1,15 @@
...
+export class SdkConflictsResolvedEvent extends DomainEvent {
+  protected readonly eventName = SdkConflictsResolvedEvent.name;
+  private static readonly message = "SDK merge conflicts resolved." as const;
+  private static readonly commandName = "sdk:generate" as const;
+
+  constructor(flags: Record<string, unknown>) {
+    super(SdkConflictsResolvedEvent.message, SdkConflictsResolvedEvent.commandName, SdkConflictsResolvedEvent.flattenFlags(flags));
+  }
+
+  private static flattenFlags(flags: Record<string, unknown>): Record<string, unknown> {
```
Shield-Jaguar: Events class should never have any logic. they are Dtos

src/types/events/sdk-save-changes.ts
```diff
@@ -0,0 +1,15 @@
...
+export class SdkSaveChangesEvent extends DomainEvent {
+  protected readonly eventName = SdkSaveChangesEvent.name;
+  private static readonly message = "Save changes completed." as const;
+  private static readonly commandName = "sdk:save-changes" as const;
+
+  constructor(flags: Record<string, unknown>) {
+    super(SdkSaveChangesEvent.message, SdkSaveChangesEvent.commandName, SdkSaveChangesEvent.flattenFlags(flags));
+  }
+
+  private static flattenFlags(flags: Record<string, unknown>): Record<string, unknown> {
```
Shield-Jaguar: Events class should never have any logic. they are Dtos

src/types/events/sdk-track-changes.ts
```diff
@@ -0,0 +1,15 @@
...
+export class SdkTrackChangesEvent extends DomainEvent {
+  protected readonly eventName = SdkTrackChangesEvent.name;
+  private static readonly message = "SDK generated with track changes enabled." as const;
+  private static readonly commandName = "sdk:generate" as const;
+
+  constructor(flags: Record<string, unknown>) {
+    super(SdkTrackChangesEvent.message, SdkTrackChangesEvent.commandName, SdkTrackChangesEvent.flattenFlags(flags));
+  }
+
+  private static flattenFlags(flags: Record<string, unknown>): Record<string, unknown> {
```
Shield-Jaguar: Events class should never have any logic. they are Dtos

src/commands/quickstart.ts
```diff
@@ -27,7 +27,8 @@ export default class Quickstart extends Command {
 
     intro("Quickstart");
     const action = new QuickstartAction(this.getConfigDir(), commandMetadata);
-    const result = await action.execute();
+    const workingDirectory = new DirectoryPath(process.cwd());
```
Shield-Jaguar: Is this a hack?

Shield-Jaguar: This doesn't look right. 
SDK generate action is not taking this parameter in input. 

The intent is not clear why this flag is passed.

src/hooks/not-found.ts
```diff
@@ -46,7 +46,7 @@ const hook: Hook.CommandNotFound = async function (opts) {
 
   if (response) {
     const confirmedSuggestion = suggestion!;
```
Shield-Jaguar, line 48: Check with @sohail2721, if he approves this change.

src/actions/portal/generate.ts
```diff
@@ -59,7 +60,11 @@ export class GenerateAction {
       if (response.isErr()) {
         const error = response.error;
         if (error instanceof ServiceError) {
-          this.prompts.portalGenerationServiceError(error);
+          if (error.code === ServiceErrorCode.SdkMergeError) {
+            this.prompts.portalGenerationSdkMergeError(error, workingDirectory);
```
Shield-Jaguar: this is the most expensive check, just for prompt you have passed the working directory from the presentation layer to this layer.

Shield-Jaguar (changes requested):

src/actions/portal/quickstart.ts
```diff
@@ -43,7 +43,7 @@ export class PortalQuickstartAction {
     this.validationService = new ValidationService(this.configDir);
   }
 
-  public readonly execute = async (): Promise<ActionResult> => {
+  public readonly execute = async (workingDirectory: DirectoryPath): Promise<ActionResult> => {
```
Shield-Jaguar: remove workingDirectory dependency

src/actions/portal/serve.ts
```diff
@@ -28,6 +30,7 @@ export class PortalServeAction {
   }
 
   public async execute(
+    workingDirectory: DirectoryPath,
     buildDirectory: DirectoryPath,
```
Shield-Jaguar, line 31: remove workingDirectory dependency

src/hooks/not-found.ts
sohail2721: With the previous logic it treated extra words as arguments i.e. PS C:\Users\SohailShahbaz\Desktop\ai-chatbot-api> apimatic auth logn --auth-key
 »   Warning: auth logn is not a apimatic command.
Did you mean auth login? (Y/n) y
 »   Error: Flag --auth-key expects a value

The new code will lose the flag information. This is a behavioral change.

src/actions/sdk/merge-source-tree.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Language } from "../../types/sdk/generate.js";
+import isInCi from "is-in-ci";
+
+export type MergeResult =
+  | { status: "success"; changesTracked: boolean }
+  | { status: "failed" }
+  | { status: "cancelled" };
+
+export class MergeSourceTreeAction {
+  private readonly prompts = new ResolveConflictsPrompts();
```
Shield-Jaguar: prompt name should match the action

src/prompts/sdk/merge-source-tree.ts
```diff
@@ -0,0 +1,79 @@
+import { isCancel, log, text } from "@clack/prompts";
+import { format as f, getTree, LeafNode, TreeNode } from "../format.js";
+import { DirectoryPath } from "../../types/file/directoryPath.js";
```
Shield-Jaguar, line 3: Make prompt classes dummy. they should only contain logci related to prompting

src/prompts/sdk/save-changes.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    return !isCancel(result);
+  }
+
+  public changesSaved() {
+    log.success("Changes saved successfully!");
+  }
+
+  private buildFileTree(
+    language: string,
```
Shield-Jaguar: prompt should not contain any logic

src/types/sdk-context.ts
```diff
@@ -9,14 +9,22 @@ export class SdkContext {
...
+    private readonly version?: string
+  ) {
   }
 
   private get zipPath(): FilePath {
     return new FilePath(this.sdkLanguageDirectory, new FileName(`${this.language}.zip`));
   }
   
   public get sdkLanguageDirectory(): DirectoryPath {
+    if (this.version) {
```
Shield-Jaguar: analyze the usage of Version. if you have passed version to this class then everything related to version should be handled in this class.

Shield-Jaguar: REview the logic in Generate.ts class where this flag is generated.

src/types/spec-context.ts
```diff
@@ -13,6 +13,14 @@ export class SpecContext {
     this.specDirectory = specDirectory;
   }
 
+  public static fromBuildDirectory(buildDirectory: DirectoryPath): SpecContext {
```
Shield-Jaguar: I think that build Context should have this method. As build directory knows the path to spec and not the spec-context.

src/types/temp-context.ts
```diff
@@ -27,4 +27,8 @@ export class TempContext {
     await this.fileService.writeFile(tempFile, portalStream);
     return tempFile;
   }
+
+  public getTempDirectory(): DirectoryPath {
```
Shield-Jaguar: this method is unused?

Shield-Jaguar (changes requested):

src/types/sdk-context.ts
Ayeshas09: The version is being resolved in VersionedBuildContext and SdkContext only determines the SDK destination based on the version. There is no additional version specific logic to move into SdkContext.

src/infrastructure/tmp-extensions.ts
Ayeshas09: This is done to handle and edge case. During conflict resolution, if the user doesn’t have VS Code, we prompt them to resolve conflicts manually and press enter. In that scenario, if the SDK directory is still open, the OS can lock the directory, which prevents tmpDir.cleanup() from deleting it.

src/hooks/not-found.ts
Ayeshas09: This behavior is still the same

$ apimatic auth logn --auth-key
 »   Warning: auth logn is not a apimatic command.
Did you mean auth login? (Y/n) y
 »   Error: Flag --auth-key expects a value
 
The change only affects how extra words from mistyped commands are handled.
Previously, inputs like sdk save changes would treat changes as an argument, leading to “Unexpected argument” errors. Now, we default argv to an empty array, so such unintended arguments are ignored and the correct validation errors (e.g., missing required flags) are shown instead.

sohail2721: Flags like --auth-key are unaffected since they come from opts.argv.

However, this is still a behavioural change: previously, extra segments from mistyped commands were treated as arguments, whereas now they’re ignored.

So, flags/explicit args remain unchanged and extra words from command are no longer passed as args.

This makes behaviour safer and more predictable, but we do lose implicit argument recovery from mistyped commands. You can discuss this with Mujahid bhai about the behaviour to keep.

src/actions/portal/generate.ts
```diff
@@ -59,7 +59,11 @@ export class GenerateAction {
       if (response.isErr()) {
         const error = response.error;
         if (error instanceof ServiceError) {
-          this.prompts.portalGenerationServiceError(error);
+          if (error.code === ServiceErrorCode.SdkMergeError) {
```
Shield-Jaguar: Suggestion: 
I usually keep if/else logic in the Application layer (as in this example), but this scenario is pushing me toward moving it into the prompt service.

Shield-Jaguar (changes requested):

src/actions/portal/generate.ts
asadali214: done

src/infrastructure/tmp-extensions.ts
Ayeshas09: We have created a seperate function now and used it only where required.

package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@apimatic/cli",
     "description": "The official CLI for APIMatic.",
-    "version": "1.1.0-beta.10",
+    "version": "1.1.0-beta.10-test",
```
Ayeshas09, line 4: @aliasghar98 We need to chnage this before merging this PR

src/hooks/not-found.ts
Shield-Jaguar: @Ayeshas09 why do we need this change?

src/actions/sdk/merge-source-tree.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      conflictedFilePaths = await this.mergeSourceTree.getConflicts(sdkDir);
+
+      if (conflictedFilePaths.length == 0) break;
+      this.prompts.conflictsStillPresent(language, conflictedFilePaths);
+    } while (true);
+
+    this.prompts.conflictsResolved(language);
+
+    const telemetryService = new TelemetryService(configDir);
```
Shield-Jaguar: TelemetryService can only exist in Command class.

src/prompts/format.ts
```diff
@@ -126,4 +131,38 @@ export function getTree(
   return output;
 }
 
+export function buildFilePathTree(rootName: string, entries: PathTreeEntry[]): string {
```
Shield-Jaguar: Review and fix. this belongs to directory or DirectoryPath Class

src/types/file/filePath.ts
```diff
@@ -15,6 +15,10 @@ export class FilePath {
     return new FilePath(newDirectory, this.fileName);
   }
 
+  public getFileName(): string {
```
Shield-Jaguar: this is exposing encapsulation

src/actions/sdk/merge-source-tree.ts
```diff
@@ -0,0 +1,80 @@
...
+import isInCi from "is-in-ci";
+import { dirPath } from "../../infrastructure/tmp-extensions.js";
+import { MergeSourceTreeContext } from "../../types/merge-source-tree-context.js";
+
+export class MergeSourceTreeAction {
+  private readonly prompts = new MergeSourceTreePrompts();
+  private readonly launcherService = new LauncherService();
+
+  public async execute(
+    mergeSourceTreeContext: MergeSourceTreeContext,
```
Shield-Jaguar: you cannot pass a context as parameter of execute method

src/actions/sdk/merge-source-tree.ts
```diff
@@ -0,0 +1,80 @@
...
+    sdkDir: DirectoryPath,
+    language: Language,
+    flags: Record<string, unknown>,
```
Shield-Jaguar: an action class cannot contain Flags.

src/actions/sdk/merge-source-tree.ts
Shield-Jaguar: The most immediate smell is the redundancy of parameters. The execute method takes mergeSourceTreeContext, yet it also explicitly takes sdkDir as a separate parameter (line 19).
•
Looking at MergeSourceTreeContext (in src/types/merge-source-tree-context.ts), it already holds sdkDir as a private property.
•
In GenerateAction.ts (the caller), both the context and the execute method are initialized with the same tempSdkDir. This duplication makes the API more brittle and harder to maintain.

src/infrastructure/launcher-service.ts
```diff
@@ -8,10 +8,57 @@ import isInCi from "is-in-ci";
 import { DirectoryPath } from "../types/file/directoryPath.js";
 
 export class LauncherService {
-  public async openFolderInIde(directoryPath: DirectoryPath, fileToOpen: FilePath): Promise<boolean> {
+  public async openFolderInIde(directoryPath: DirectoryPath, ...filesToOpen: FilePath[]): Promise<boolean> {
```
Shield-Jaguar: what if the user doesn't have the code? what will happen?

src/infrastructure/launcher-service.ts
```diff
@@ -8,10 +8,57 @@ import isInCi from "is-in-ci";
...
-      await execa("code", [directoryPath.toString(), fileToOpen.toString()]);
+      const args = [directoryPath.toString(), ...filesToOpen.map(f => f.toString())];
+      await execa("code", args);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  public async openDiffsInSourceControl(
```
Shield-Jaguar: this method doesn't belong to launcher service.  launcher service is infra service and its dummy. it should not contain any logic.

src/infrastructure/tmp-extensions.ts
```diff
@@ -1,8 +1,31 @@
...
+import { FileService } from "./file-service.js";
 
 export function withDirPath<T>(
   fn: (results: DirectoryPath) => Promise<T>,
 ): Promise<T> {
   return withDir(results => fn(new DirectoryPath(results.path)), { unsafeCleanup: true });
 }
+
+export async function dirPath<T>(
+  originalDirPath: DirectoryPath,
```
Shield-Jaguar: this method is not and extension method. there is complete logic living in this only possibility is and extension method `withdir` 

this service cannot contain any other service. like file service tec.

Shield-Jaguar (changes requested):

src/infrastructure/launcher-service.ts
asadali214: Then this function will return false, and we will fall back to the manual review flow for the user

src/infrastructure/tmp-extensions.ts
asadali214: Removed it

src/actions/sdk/merge-source-tree.ts
asadali214: removed flags, these were passed for telemetry events

src/actions/sdk/merge-source-tree.ts
asadali214: This is sorted, the action will create its context itself now

src/actions/sdk/merge-source-tree.ts
asadali214: This is fixed^

CODEBASE_CONTEXT.md
```diff
@@ -0,0 +1,1152 @@
+# APIMatic CLI — Codebase Context & Architecture Guide
```
Shield-Jaguar: This file can be deleted now

src/types/sdk-input-context.ts
```diff
@@ -0,0 +1,109 @@
...
+import { FilePath } from "./file/filePath.js";
+import { Language } from "./sdk/generate.js";
+
+export class SdkInputContext {
+  private readonly fileService = new FileService();
+  private readonly zipService = new ZipService();
+  private readonly gitService = new GitService();
+
+  constructor(
+    private readonly sdkDirectoryInput: string | undefined,
```
Shield-Jaguar: context files cannot contain string values.

src/types/sdk-input-context.ts
```diff
@@ -0,0 +1,109 @@
...
+    updatedSdkDirectory: DirectoryPath,
+    sourceTreePath: FilePath
+  ): Promise<void> {
+    const sdkGitDir = updatedSdkDirectory.join(".git");
+    await this.gitService.commitReviewedChanges(updatedSdkDirectory);
+    await this.gitService.forceCheckoutMainBranch(updatedSdkDirectory);
+    await this.zipService.archive(sdkGitDir, sourceTreePath);
+  }
+
+  public async classifyChangedFiles(
```
Shield-Jaguar: This methods read refactoring. 

the input and returnt type makes no scense. 

GitFileStatus is infra level dependncy. it should never be in the application layer or domain layer.

Shield-Jaguar (changes requested):

src/types/file/directoryPath.ts
```diff
@@ -28,7 +29,44 @@ export class DirectoryPath {
   public isEqual(other: DirectoryPath) {
     return this.directoryPath === other.directoryPath;
   }
+
   public leafName() {
     return path.basename(this.directoryPath);
   }
+
+  public toTreeNode(files: {
```
Shield-Jaguar: this is not the right place for to tree path. This is using leafname which is already a violation.

Shield-Jaguar (changes requested):

src/infrastructure/service-error.ts
```diff
@@ -1,21 +1,35 @@
...
 
+export enum ServiceErrorCode {
+  NotFound = "NOT_FOUND",
+  ServerError = "SERVER_ERROR",
+  NetworkError = "NETWORK_ERROR",
+  InvalidResponse = "INVALID_RESPONSE",
+  UnAuthorized = "UNAUTHORIZED",
+  BadRequest = "BAD_REQUEST",
+  Forbidden = "FORBIDDEN",
+  SdkMergeError = "SDK_MERGE_ERROR"
```
Shield-Jaguar: you cannot return SDK_MERGE_ERROR. because this is your custom logic. 

the if condition used for SDK_MERGE_ERROR is your application logic. not infrastructure logic..

this logic should be moved one layer up. Remove SdkMergeError. 

ideally you should not export a type from infrastructure layer. 

because dependencies goes upward.

 Side effects of the current design:
  - ServiceErrorCode is now a dependency of the Prompts layer — infra leaking up two layers
  - portalGenerationServiceError has mixed responsibilities: it both logs the error and conditionally renders next-steps UI
  - If a new error code needs special UI treatment, you have to modify the Prompts layer for what is really an Action-level routing
  decision

Shield-Jaguar (changes requested):

src/infrastructure/file-service.ts
```diff
@@ -41,6 +42,20 @@ export class FileService {
     await fsExtra.emptyDir(dir.toString()); // removes everything inside, keeps the dir
   }
 
+  public async cleanDirectoryExcluding(dir: DirectoryPath, excludeNames: string[]): Promise<void> {
```
Shield-Jaguar: exclude can only be FilePaths[] or FileNames[]. it cannot be string

src/infrastructure/file-service.ts
```diff
@@ -79,6 +102,33 @@ export class FileService {
     );
   }
 
+  public async copyDirectoryExcluding(
+    source: DirectoryPath,
+    destination: DirectoryPath,
+    excludeNames: string[]
```
Shield-Jaguar: exclude names cannot be strings[]. it has to be filepath or FileName etc

src/infrastructure/file-service.ts
```diff
@@ -149,6 +215,32 @@ export class FileService {
...
+  }
+
+  public async normalizeFileLineEndings(filePath: FilePath): Promise<void> {
+    if (process.platform !== "win32") return;
+    const content = await this.readFile(filePath);
+    const normalizedContent = content.replace(/\r\n/g, "\n").replace(/\r/g, "\n");
+    await this.writeContents(filePath, normalizedContent);
+  }
+
+  public async filterFilesWithConflictMarkers(updatedFiles: FilePath[]): Promise<FilePath[]> {
```
Shield-Jaguar: this is custom logic moved to file serivce. this has hidden domain logic or logic related to git

src/infrastructure/file-service.ts
```diff
@@ -93,6 +143,18 @@ export class FileService {
     }
   }
 
+  public async forceDeleteDirectory(dirPath: DirectoryPath, shouldRetry: () => Promise<boolean>): Promise<boolean> {
```
Shield-Jaguar: FileService is infrastructure — it wraps filesystem operations (read, write, copy, delete). forceDeleteDirectory is not a filesystem
  operation; it is retry orchestration with embedded policy. Specifically:

  1. Wrong layer — Retry logic with timing (Date.now(), setTimeout), a deadline window, and a user-decision callback (shouldRetry: () =>
   Promise<boolean>) belongs in the Action or Application layer, not in infrastructure. Infrastructure services should be stateless,
  side-effect-free wrappers around external systems.
  2. shouldRetry callback leaks interaction concerns into infrastructure — That callback exists to ask the user a question ("try
  again?"). Prompting or delegating UI decisions is the Prompts/Action layer's responsibility. Infrastructure should never accept
  callbacks that touch user interaction.
  3. The underlying capability already exists — deleteDirectory (line 139) already does the actual work. The retry wrapper is pure
  orchestration on top of it — it adds no new filesystem capability.
  4. It encodes policy that varies by caller — The 2-second deadline and 50ms poll interval are arbitrary choices that different callers
   might reasonably want to differ. Baking them into a shared infrastructure service locks all callers into one undocumented policy.

  Suggested resolution for the PR: Remove forceDeleteDirectory from FileService.

src/prompts/portal/generate.ts
```diff
@@ -47,6 +47,12 @@ export class PortalGeneratePrompts {
 
   public portalGenerationServiceError(serviceError: ServiceError) {
     log.error(serviceError.errorMessage);
+    if (serviceError.code === ServiceErrorCode.SdkMergeError) {
```
Shield-Jaguar: this if condition is hidden deep into prompt service. this is incorrect.

this flag condition should not exit at all. this is not a real service Error. but a reaction to a a response received from error.

Shield-Jaguar (changes requested):

src/types/sdk-context.ts
```diff
@@ -2,35 +2,61 @@ import { FileService } from "../infrastructure/file-service.js";
...
 
-  constructor(private readonly sdkDirectory: DirectoryPath, private readonly language: Language) {
+  constructor(
+    private readonly sdkDirectory: DirectoryPath,
+    private readonly language: Language,
+    private readonly requireUncustomizedDir: boolean,
+    private readonly version?: string
+  ) {  }
+
+  public getSdkLanguageDirectory(): DirectoryPath {
```
Shield-Jaguar: Rule: "DON'T expose internal paths as public getters — no public get outputDirectory() or public get filePath()."

src/types/sdk-context.ts
```diff
@@ -2,35 +2,61 @@ import { FileService } from "../infrastructure/file-service.js";
...
+    await this.zipService.unArchive(tempSdkSourceTree, gitSourceTreeDir);
   }
 
-  public async save(tempPortalFilePath: FilePath, zipPortal: boolean) {
-    await this.fileService.cleanDirectory(this.sdkLanguageDirectory);
-    if (zipPortal) {
-      await this.fileService.copy(tempPortalFilePath, this.zipPath);
-    } else {
-      await this.zipService.unArchive(tempPortalFilePath, this.sdkLanguageDirectory);
+  public async save(tempSdkDirectory: DirectoryPath, zipSdk: boolean) : Promise<DirectoryPath> {
```
Shield-Jaguar: what is the point here? we only return directory after save. because this is when sdk is saved.  you are returning sdkLanguageDir which is already exposed with getSdkLanguageDirectory()

this violates the context classes principles. 

we use the return of called function in next function. This is how context classes are designed.

Shield-Jaguar (changes requested):

src/types/sdk-context.ts
asadali214: Removed^

src/types/sdk-context.ts
asadali214: I have incorporated this feedback

src/infrastructure/file-service.ts
asadali214: Done

src/infrastructure/file-service.ts
asadali214: Changed to FileName[]

src/infrastructure/file-service.ts
asadali214: changed to FileName[]

src/types/sdk-input-context.ts
asadali214: changed to DirectoryPath

src/types/file/directoryPath.ts
asadali214: Removed from here

src/types/sdk-input-context.ts
asadali214: This class has been renamed to save-changes-context and totally refactored

src/infrastructure/file-service.ts
asadali214: This is fixed, we are now polling until the directory is remains open in any application^

src/prompts/portal/generate.ts
asadali214: This is fixed

src/infrastructure/service-error.ts
asadali214: The SDK_MERGE_ERROR code is removed and the issue is handled using the problem detail error addition in ServiceError

src/infrastructure/git-service.ts
```diff
@@ -0,0 +1,128 @@
...
+      .filter(([, headStatus, workdirStatus, stageStatus]) => headStatus !== workdirStatus || stageStatus === 3)
+      .map(([relativePath, workdirStatus, headStatus]) => {
+        if (headStatus === 0) {
+          return { relativePath, description: "# Added" };
+        }
+        if (workdirStatus === 0) {
+          return { relativePath, description: "# Deleted" };
+        }
+        return { relativePath, description: "# Modified" };
+      });
```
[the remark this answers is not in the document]
asadali214: fixed

src/infrastructure/git-service.ts
```diff
@@ -0,0 +1,128 @@
...
+      statusMatrix
+        // include updated and resolved conflict files
+        .filter(([, headStatus, workdirStatus, stageStatus]) => headStatus !== workdirStatus || stageStatus === 3)
+        .map(([relativePath, workdirStatus]) => {
+          if (workdirStatus === 0) {
+            return git.remove({ fs, dir: dir.toString(), filepath: relativePath });
+          }
+          return git.add({ fs, dir: dir.toString(), filepath: relativePath });
+        })
+    );
```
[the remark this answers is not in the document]
asadali214: fixed

src/actions/sdk/generate.ts
```diff
@@ -25,79 +26,117 @@ export class GenerateAction {
...
+        this.prompts.sdkGenerated(await sdkContext.save(tempSdk, zipSdk));
+        return ActionResult.success();
+      }
 
-      this.prompts.sdkGenerated(sdkLanguageDirectory);
+      const tempSdkWithSourceTree = await sdkContext.loadSdkWithSourceTreeInTempDirectory(
+        tempDirectory,
+        await tempContext.save(response.value.sdk),
+        await tempContext.save(response.value.sdkSourceTree)
+      );
```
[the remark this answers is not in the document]
asadali214: fixed

Shield-Jaguar (approved):
There are still some fundamental code issues that needs to be fixed. They will be fixed later in another pull request.

sohail2721 (approved):

asadali214 (approved):

PR #297, opened 2026-07-14
feat: surface API 401 error message with login suggestion in portal generation

MuhammadRafay1:
## Problem

Running portal commands without a valid session (e.g. `apimatic portal serve` while logged out) showed the generic fallback:

> An unexpected error occurred, please try again later. If the problem persists, please reach out to our team at support@apimatic.io

even though the API returns a clear reason (`{"message": "Authorization has been denied for this request."}`). The SDK throws its typed `UnauthorizedResponseError` for 401s, but `handleServiceError` only recognizes axios errors, so the  generic `ServerError`.

## Changes

- **`portal-service.ts`**: In `generatePortal`'s catch, handle `ApiError` with status 401 — extract the API's message from the deserialized error `result` and return a `ServiceError` that pairs it with a suggested fix. All other errors continue to flow through `handleServiceError` unchanged (kept scoped on purpose; message handling differs per flow).

- **`service-error.ts`**: New `ServiceError.unauthorized(apiMessage)` factory (alongside `badRequest`/`forbidden`/`notFound`) that appends "run `apimatic auth login` or provide a valid auth key using the `--auth-key` flag" to the API's message.

## Result

■  Portal Generation failed.
■  Authorization has been denied for this request. Please
   to log in, or provide a valid auth key using the --auth-key flag.

## Testing

- Reproduced end-to-end: `portal serve` with an empty config dir (logged out) against the production API — enhanced message shown.
- Verified the axios 401 path is unaffected: `auth status` with an invalid stored key still prints "Invalid API key provided." (relies on the `ServiceError.UnAuthorized` singleton identity check, which is preserved).
- `pnpm test`: 45 passing, lint clean.

### DIFF
--- a/src/infrastructure/service-error.ts
+++ b/src/infrastructure/service-error.ts
@@ -1,4 +1,5 @@
 import axios from "axios";
+import { ApiError } from "@apimatic/sdk";
 import { format as f } from "../prompts/format.js";
 
 export enum ServiceErrorCode {
@@ -63,6 +64,15 @@ export class ServiceError {
 }
 
 export function handleServiceError(error: unknown): ServiceError {
+  // SDK controllers throw typed `ApiError`s (not axios errors). The API reports
+  // the reason as {"message": "..."} deserialized into `result`.
+  if (error instanceof ApiError) {
+    const apiMessage = (error.result as { message?: string } | undefined)?.message ?? null;
+    if (error.statusCode === 401) return ServiceError.unauthorized(apiMessage);
+    if (error.statusCode === 404) return ServiceError.NotFound;
+    if (error.statusCode === 500) return ServiceError.ServerError;
+  }
+
   if (axios.isAxiosError(error)) {
     const status = error.response?.status;
     if (status === 401) return ServiceError.UnAuthorized;
--- a/src/infrastructure/service-error.ts
+++ b/src/infrastructure/service-error.ts
@@ -63,15 +63,19 @@ export class ServiceError {
   }
 }
 
-export function handleServiceError(error: unknown): ServiceError {
-  // SDK controllers throw typed `ApiError`s (not axios errors). The API reports
-  // the reason as {"message": "..."} deserialized into `result`.
-  if (error instanceof ApiError) {
+// SDK controllers throw typed `ApiError`s (not axios errors). The API reports
+// the reason as {"message": "..."} deserialized into `result`.
+function mapApiError(error: ApiError): ServiceError {
+  if (error.statusCode === 401) {
     const apiMessage = (error.result as { message?: string } | undefined)?.message ?? null;
-    if (error.statusCode === 401) return ServiceError.unauthorized(apiMessage);
-    if (error.statusCode === 404) return ServiceError.NotFound;
-    if (error.statusCode === 500) return ServiceError.ServerError;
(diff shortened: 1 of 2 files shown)


sohail2721 (approved):
LGTM

src/infrastructure/service-error.ts
```diff
@@ -54,7 +63,36 @@ export class ServiceError {
...
+  if (error.statusCode === 401) {
+    const apiMessage = (error.result as { message?: string } | undefined)?.message ?? null;
+    return ServiceError.unauthorized(apiMessage);
+  }
+  if (error.statusCode === 404) return ServiceError.NotFound;
+  return ServiceError.ServerError;
+}
+
 export function handleServiceError(error: unknown): ServiceError {
+  if (error instanceof ApiError) return mapApiError(error);
```
mehnoorsiddiqui, line 97: handleServiceError is shared by 7 services (api-, auth-, file-download-, publishing-api-, validation-, portal-service). The new if (error instanceof ApiError) branch changes behavior for every SDK-based caller, not just portal.Is this intentional

src/infrastructure/service-error.ts
```diff
@@ -30,6 +31,14 @@ export class ServiceError {
   static notFound(customMessage: string): ServiceError {
     return new ServiceError(ServiceErrorCode.NotFound, customMessage, {});
   }
+  static unauthorized(apiMessage: string | null): ServiceError {
```
mehnoorsiddiqui: ServiceError.UnAuthorized (singleton, generic message) vs ServiceError.unauthorized() (factory, enhanced message) differ only by casing. Both yield ServiceErrorCode.UnAuthorized. This is easy to mistype and produces two silently different behaviors. Consider a more distinct name (e.g. unauthorizedWithHint).

src/infrastructure/service-error.ts
```diff
@@ -54,7 +63,36 @@ export class ServiceError {
   }
 }
 
+// A ProblemDetails body carries a `title` plus a map of field errors; surface
+// the title and the first field message.
+function mapProblemDetailsError(error: ProblemDetailsError): ServiceError | null {
+  // TODO: This only picks the first error message, improve it to show all errors.
+  const errors = error.result!.errors as Record<string, string[]>;
```
mehnoorsiddiqui: the Object.values(error.result!.errors) call crashes with a TypeError when the optional errors field is absent — previously buried in portal-service's catch blocks, now centralized into mapProblemDetailsError where it runs ahead of the 401 branch, so it's worth hardening (error.result?.errors ?? {}) as part of this refactor.

mehnoorsiddiqui:
```suggestion
const errors = (error.result?.errors ?? {}) as Record<string, string[]>;
```

src/infrastructure/service-error.ts
MuhammadRafay1: In practice portal-service is the only caller that routes an SDK ApiError through it today, every other service (api, auth, file-download, publishing-api, validation) is axios-based and still hits the axios branch unchanged, and transformation-service uses its own handler and never calls handleServiceError. So there's no behavior change for non-portal callers right now.
In the first commit the error handling was done inside portal-service. The refactor commits later on moved these to error-service

Issue #301, opened 2026-07-28
portal toc new: unauthenticated run writes a stub toc.yml with exit 0, then hangs


MuhammadRafay1:
## Summary

`apimatic portal toc new` handles a failed TOC extraction badly in two independent
ways. Reproduce with `apimatic auth logout` then
`apimatic portal toc new --input <dir-with-src/spec> --force`:

▲  Falling back to the default TOC structure.
●  The TOC file successfully created at: …\src\content\toc.yml
└  Succeeded
<hangs — never exits>

**1. The 401 is swallowed.** `actions/portal/toc/new-toc.ts:86-89` treats every
`generateTocData` error as benign: it overwrites the existing `toc.yml` with one
containing no endpoint/model/webhook components and returns
`ActionResult.success()` (exit 0). The user is never told to log in, and CI sees
success. `portal-service.ts:333` already produces the right message — the action
discards it.

**2. The process then hangs indefinitely.** `generateTocData` is a
stream-response endpoint, so on error the SDK returns `ApiError.body` as an
undrained `IncomingMessage`. The TLS socket stays open, and since `outro()` only
sets `process.exitCode` (`prompts/format.ts:48`) with no `process.exit` anywhere,
Node waits forever. Confirmed by handle probe (`TLSSocket … readable=true`);
destroying that stream turns a 90s hang into a 3s exit 0. **Not specific to this
command or to being logged out** — latent in any stream-endpoint error, including
the portal and SDK download calls.

## Expected

An actionable auth message, a non-zero exit, the existing `toc.yml` left
untouched, and the process exiting.

MuhammadRafay1:
PR for this issue: https://github.com/apimatic/apimatic-cli/pull/303

MuhammadRafay1:
Fixed in release: Alpha 1.2.1

PR #303, opened 2026-08-03
fix: surface TOC extraction failures instead of writing a partial toc…

MuhammadRafay1:
## The [issue](https://github.com/apimatic/apimatic-cli/issues/301)

`portal toc new` treated *every* failure to extract endpoint data as a reason to fall back to the default TOC:

```ts
if (result.isErr()) {
  this.prompts.fallingBackToDefault();
  return TocComponents.empty();
}
```

`TocComponents.empty()` means "no Endpoints, Events or Models sections". So when a user was logged out — or hit a network error, an expired key, or an invalid spec — the command would:

1. Write a `toc.yml` containing only custom content, **overwriting a possibly valid existing TOC**,
2. Report success, giving no indication that the endpoint sections were dropped, and
3. **Hang** instead of returning to the shell.

### 1. Propagate the error instead of silently degrading

`actions/portal/toc/new-toc.ts`

The extraction block now returns `Result<TocComponents, ServiceError>`. A failed extraction returns `err(...)`; the caller reports the reason (auth, network, invalid spec) and returns `ActionResult.failed()`, leaving `toc.yml` untouched. `specFileStream.close()` moved into a `finally` so it is released on both paths.

### 2. Require a `spec/` directory

`actions/portal/toc/new-toc.ts`, `prompts/portal/toc/new-toc.ts`

The spec is what the TOC is built from, and the next step — portal generation — needs it regardless, so a TOC generated without one only defers the failure to a later command. The default-TOC fallback is dropped entirely: the command now fails when the spec is unusable, whether or not any `--expand-*` flag was passed.

Missing and empty are reported separately. `SpecContext.validate()` treats both as invalid, but only one of them is fixed by adding files to a directory that already exists. Both messages name the build directory, since that is the path the user passed.

Auth is deliberately not part of this check: with no spec no request is made, so an unauthorized message here would name a cause that did not occur. A logged-out user is blocked on the spec first, and on auth once a spec exists.

This removes `TocComponents.empty()` and the `fallingBackToDefault` prompt, both now unreachable.

### 3. Release stream error bodies

`utils/utils.ts`, `infrastructure/service-error.ts`, `infrastructure/services/transformation-service.ts`

The hang had a separate cause. `generateTocData` is a `callAsStream` endpoint, so on failure the SDK hands back an undrained `IncomingMessage` on `ApiError.body`. Its socket stays open and keeps the Node event loop alive, and since `outro()` only sets `process.exitCode` — we never call `process.exit()` — the process printed its outro and then sat there.

This is narrower than "stream endpoints leak their error bodies". `@apimatic/core` normally drains them itself: `loadResult` reads the body into `error.result` before the error is thrown. But it only runs on the `throwOn` path, and only `if (errorConstructor.prototype instanceof ApiError)` — strict subclasses. Two gaps follow:

- **No matching `throwOn`** — the generic response validator throws a bare `ApiError` with no `loadResult` call at all.
- **`throwOn` registered with `ApiError` itself** — fails the subclass check, so `loadResult` is skipped.

`tableOfContentsController` and `transformationController` register **zero** `throwOn`s, so every error from `generateTocData` and `downloadTransformedFile` lands in the first gap with a live stream body. That is the leak.

New `discardStreamBody(body)` in `utils/utils.ts`, alongside its counterpart `parseStreamBodyToJson`. `destroy()` is not declared on `NodeJS.ReadableStream` — it lives on `stream.Readable`/`IncomingMessage` — so narrowing the `string | Blob | NodeJS.ReadableStream` union still will not let us call it; an `isDestroyable` type guard probes for the method instead. That probe is a real platform branch, not defensiveness: the axios adapter yields a stream on Node and a Blob elsewhere.

Call sites:

- `service-error.ts` — from `handleServiceError`, covering `generateTocData`.
- `transformation-service.ts` — covering `downloadTransformedFile`.

Not applied where it would be wrong or redundant:

- **`PortalService.generatePortal`'s 422 path** returns `error.body` as the validation report and short-circuits before `handleServiceError`, so that stream is never destroyed. This is load-bearing rather than incidental: the SDK registers `throwOn(422, ApiError)` — `ApiError` itself, not a subclass — which is the second gap above, and the only reason the body arrives readable.
- **`validation-service.ts`** uses `validateApiViaFileV2`, a `callAsJson` endpoint with subclass `throwOn`s. Its body is always a string, so the call would be dead code.

### 4. Renamed prompt method

`PortalNewTocPrompts.logError` → `tocExtractionFailed`, to name what it reports.

## Tests

`test/utils/utils.test.ts` covers `discardStreamBody`. Its whole contract is a duck-type probe over a `string | Blob | NodeJS.ReadableStream` union, so the branches are the thing worth pinning: destroys a stream, leaves a string readable for the caller that parses it afterwards, ignores a Blob-like body, and stays quiet on an already-destroyed stream — the SDK drains error bodies itself on the `throwOn` path, so a finished body can reach it.

`test/actions/portal/toc/new-toc.test.ts` covers the spec requirement: missing, empty, and dotfiles-only, which `FileService.directoryEmpty` also treats as empty. All three return before any request is made, so no service stubbing is needed. A fourth asserts nothing is written to the build directory — the regression that mattered, since the old fallback overwrote a possibly valid `toc.yml` and exited 0.

Suite is green: 55 passing, 11 pending (the pre-existing skipped live-integration tests in `portal/generate.test.ts`).

## Behavior changes

Both are limited to `portal toc new`. `PortalNewTocAction` is only used by `commands/portal/toc/new.ts`, so nothing else is affected.

- **A failed extraction now fails the command** rather than writing a partial TOC — including when the user is logged out. Previously it exited 0 with a degraded file.
- **A missing or empty `spec/` directory now fails the command.** Previously a build without `spec/` produced the default TOC and exited 0.

### DIFF
--- a/src/infrastructure/services/transformation-service.ts
+++ b/src/infrastructure/services/transformation-service.ts
@@ -80,6 +80,23 @@ export class TransformationService {
     };
   };
 
+  /**
+   * A 401 body is not guaranteed to be JSON: a stream endpoint hands back an
+   * `IncomingMessage`, an empty body parses to nothing, and an authenticating
+   * proxy can answer with an HTML error page. Throwing here would escape the
+   * catch in `transformViaFile` — this runs inside the `err(await ...)` it
+   * builds — and surface as an unhandled rejection, skipping the outro and the
+   * failure telemetry. Fall back to the hint's own default message instead.
+   */
+  private parseApiMessage(body: unknown): string | null {
+    if (typeof body !== "string") return null;
+    try {
+      return (JSON.parse(body) as { message?: string })?.message ?? null;
+    } catch {
+      return null;
+    }
+  }
+
   private readonly handleTransformationErrors = async (error: unknown): Promise<string> => {
     if (error instanceof ApiError) {
       const apiError = error as ApiError;
@@ -87,8 +104,7 @@ export class TransformationService {
         if (apiError.statusCode === 400) {
           return "Your API Definition is invalid. Please use the APIMatic VS Code Extension to fix the errors and try again.";
         } else if (apiError.statusCode === 401) {
-          const message = JSON.parse(apiError.body as string).message;
-          return ServiceError.unauthorizedWithHint(message).errorMessage;
+          return ServiceError.unauthorizedWithHint(this.parseApiMessage(apiError.body)).errorMessage;
         }
         return `Error ${apiError.statusCode}: An error occurred during the transformation. Please try again or contact support@apimatic.io for assistance.`;
       } finally {
--- a/src/infrastructure/services/transformation-service.ts
+++ b/src/infrastructure/services/transformation-service.ts
@@ -17,7 +17,7 @@ import { apiClientFactory } from "./api-client-factory.js";
 import { FilePath } from "../../types/file/filePath.js";
 import { CommandMetadata } from "../../types/common/command-metadata.js";
 import { err, ok, Result} from "neverthrow";
-import { ServiceError } from "../service-error.js";
+import { discardStreamBody, ServiceError } from "../service-error.js";
 
 export interface TransformViaFileParams {
   file: FilePath;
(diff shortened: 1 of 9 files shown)


src/infrastructure/services/transformation-service.ts
```diff
@@ -83,13 +83,20 @@ export class TransformationService {
...
-      if (apiError.statusCode === 400) {
-        return "Your API Definition is invalid. Please use the APIMatic VS Code Extension to fix the errors and try again.";
-      } else if (apiError.statusCode === 401) {
-        const message = JSON.parse(apiError.body as string).message;
-        return ServiceError.unauthorizedWithHint(message).errorMessage;
+      try {
+        if (apiError.statusCode === 400) {
+          return "Your API Definition is invalid. Please use the APIMatic VS Code Extension to fix the errors and try again.";
+        } else if (apiError.statusCode === 401) {
+          const message = JSON.parse(apiError.body as string).message;
```
mehnoorsiddiqui: ### Unguarded `JSON.parse` on an Error Path

The following line performs an unguarded JSON parse on an error path:

```ts
const message = JSON.parse(apiError.body as string).message;
```

There are three common scenarios where this throws:

| Input | Result |
|--------|--------|
| Stream body (401 from `downloadTransformedFile`, a `callAsStream` endpoint) | `JSON.parse` stringifies the `IncomingMessage` to `"[object Object]"` → `SyntaxError` (confirmed) |
| Empty body | `JSON.parse("")` → `SyntaxError` |
| Non-JSON body (gateway/proxy HTML, e.g. a corporate MITM 401 page) | `SyntaxError` |

The third case is the realistic one. Enterprise users behind an authenticating proxy often receive an HTML error page instead of JSON.

#### Consequence

The `SyntaxError` escapes `handleTransformationErrors`, then escapes the `catch` block in `transformViaFile` (because the throw occurs inside `err(await this.handleTransformationErrors(error))`), and ultimately surfaces as an unhandled rejection.

As a result:

- Users receive a raw stack trace instead of the intended error message.
- `outro(result)` is skipped.
- The process exits with the wrong exit code.
- Failure telemetry is not recorded correctly.

#### Severity

**Medium** — this does not cause data loss or corruption, but it degrades an already-failing path into an unhandled exception and breaks expected error reporting.

#### Why This Belongs in This PR

Although this issue is pre-existing, this PR modifies these exact lines by wrapping them in `try/finally` and adds a comment asserting that this branch safely reads the body.

Additionally:

- The safe-parse pattern already exists in the sibling file (`validation-service.ts`, lines 225–234).
- `unauthorizedWithHint` already accepts `string | null`.

This is a small, low-risk fix (roughly five lines) that aligns this code path with the existing implementation.

mehnoorsiddiqui (approved):

Issue #313, opened 2026-08-10
sdk publish: no telemetry for publishes that fail after being accepted


mrafnadeem-apimatic:
## Problem

`apimatic sdk publish` only reports telemetry when the publishing API **rejects** the publish request. `SdkPublishAction` calls `onPublishSdkError` in exactly one place:

https://github.com/apimatic/apimatic-cli/blob/dev/src/actions/sdk/publish.ts#L95-L99

```ts
if (publishSdkResponse.isErr()) {
  this.prompts.sdkPublishingServiceError(publishSdkResponse.error);
  onPublishSdkError(publishSdkResponse.error.errorMessage);
  return ActionResult.failed();
}
```

which fires `SdkPublishValidationFailedEvent` (400/403/404, auth, validation).

Nothing is tracked for the case where the request is **accepted** and publishing then terminates in `Failed`, `Exception`, or `InternalError`, nor for a `getSdkPublishingLog` error that ends the poll. The CLI observes these outcomes — it prints them and exits `1` — but never reports them, so the failure rate of publishes that actually reached the publishing service is invisible.

## Why it wasn't folded into the existing event

`SdkPublishValidationFailedEvent` models a rejected publish *request*, and its `message` carries a real service error string. A post-acceptance failure has no equivalent message (only per-target `eventType` values), and routing it through an event named `…ValidationFailed` would mean that metric silently stops meaning one thing.

## Suggested shape

- A dedicated event, e.g. `SdkPublishFailedEvent extends DomainEvent`, with a message summarising which targets failed and with what `eventType`.
- A second callback threaded from `commands/sdk/publish.ts` through both mode actions into `SdkPublishAction`, alongside the existing `onPublishSdkError`.
- Decide whether a status-fetch error (outcome unknown) is reported as the same event, a distinct one, or not at all — it is not a publish failure, and conflating it would inflate the failure rate.

## Context

Both interactive and non-interactive `sdk publish` now poll to a terminal state, so both modes reach this untracked failure path. Deliberately left out of the change that added polling to interactive mode, to keep that a behavioral-parity fix. Related: #312.

Issue #312, opened 2026-08-10
sdk publish: missing array length check makes an empty publish-log report a false success


mrafnadeem-apimatic:
## Problem

`SdkPublishNonInteractivePrompts.pollPublishingStatus` decides completion with `Array.prototype.every`, which is vacuously `true` on an empty array:

https://github.com/apimatic/apimatic-cli/blob/dev/src/prompts/sdk/publish/non-interactive.ts#L125-L144

```ts
const { events } = publishingLogResult.value;
const executionCompleted = events.every((event) => TERMINAL_STATES.has(event.eventType));
...
const isExecutionSuccessful = events.every((event) => event.eventType === 'Succeeded');
```

There is no length check on `events`, so if the publishing API ever returns `{ events: [] }` for a publish log, the very first poll concludes that publishing both **completed** and **succeeded**. The CLI stops the spinner with an empty status line, prints the "Next Steps" note, and exits `0` for a publish whose outcome is unknown.

A related case falls out of the same missing check: if `events` contains only *some* of the requested publish targets (e.g. the `SourceCode` event exists and already reads `Succeeded` while the `Package` event has not been created yet), `every` is satisfied over the partial set and the command reports full success while package publishing has not started. Whether this second case is reachable depends on whether the publishing API creates all event rows atomically when it accepts the publish request — worth confirming before picking a fix.

## Impact

`apimatic sdk publish` exits `0` and prints a success-shaped status line for a publish it never actually observed reaching a terminal state. This now affects interactive mode as well, since interactive polls for completion through the same shared code.

## Possible fixes

1. Guard on target coverage — completion requires an event for every requested publish type *and* all events terminal. The action already has `publishTypes`; needs a `'package' -> 'Package'` / `'sourcecode' -> 'SourceCode'` mapping, since the request-side `PublishType` enum and the log event's `publishType` use different casing.
2. Minimal: require `events.length > 0` before evaluating the two predicates. Sufficient only if the API creates all event rows atomically.
3. Treat a missing/partial event set as an explicit error and stop the spinner with a message naming the missing target(s), rather than continuing to wait.

Whichever is chosen, the failure mode should be distinguishable in the output from "publishing failed" — the CLI does not know the outcome in this state.

## Context

Deliberately deferred out of the change that makes interactive `sdk publish` poll for completion, to keep that change a behavioral-parity fix with no predicate changes. Two adjacent items also parked here:

[...]

PR #315, opened 2026-08-11
fix: poll for publish completion in interactive sdk publish

mrafnadeem-apimatic:
## Problem

The two modes of `apimatic sdk publish` disagreed about when the command was done.

- **Interactive** returned as soon as the publishing API accepted the POST. It printed *"To view the status of publishing, please visit…"* and returned `success()`, so the command exited `0` whether publishing later succeeded or failed.
- **Non-interactive** polled `getSdkPublishingLog` every 10 s to a terminal state, printed the closing note, and returned `failed()` when publishing did not succeed.

On cancellation with Ctrl+C, the CLI should say, `Publishing is still running on APIMatic and will continue without the CLI.`.

## Change

Polling moved into the shared `SdkPublishAction`, so both modes run one identical post-POST sequence: running notice → poll → closing note with the log URL. The poll runs *after* the `withDirPath` block closes, releasing the temp directory holding the zipped SDK before the CLI sits waiting for minutes.

- Interactive's `sdkPublishingInProgress` message is deleted — publishing is no longer in flight when the command exits, so its wording had become wrong. Non-interactive's log URL wording is reused verbatim for both outcomes; the log URL is useful on success and essential on failure.
- `pollPublishingStatus` returns `'succeeded' | 'failed' | 'cancelled'` (exported from the prompts module, following `QuickstartFlow`). A boolean cannot express cancel-vs-fail, and the distinction matters for planned work. `ActionResult` stays out of the prompts layer.
- `SdkPublishAction.execute` now returns `Promise<ActionResult>` rather than `ActionResult<PublishingInfo>`. Nothing reads the payload now that the action prints its own notes, and the old signature carried a hazard: the dry-run branch returns `success()` with no value, so `getValue()` on it throws.
- Non-interactive behaviour is unchanged.

## Deliberately not in scope

- **The completion predicate is unchanged.** `events.every(...)` is vacuously true on an empty array, so an empty or partially-populated publish log reports instant false success. Tracked in #312, along with the lack of retry tolerance on transient status-fetch errors and the unbounded wait.
- **No timeout.** Any ceiling is a guess about the slowest legitimate publish, and guessing low turns a slow success into a reported failure. Ctrl+C is the interactive escape hatch; job timeouts are the CI one.
- **The poll loop stays in the prompts layer**, moved verbatim. Relocating it to a service (as `portal-service.ts`'s `pollUntilCompleted` would suggest) means inventing a status-callback seam across a layer boundary — a refactor riding along on a parity fix.
- **No telemetry** for publishes that fail *after* being accepted. `SdkPublishValidationFailedEvent` models a rejected publish *request*; folding remote failures into it would corrupt that metric. Tracked in #313.

## Testing

`test/` has no publish coverage and no prompts tests at all. Meaningful coverage here needs either fake timers plus stdout capture against a live clack spinner, or dependency injection into `SdkPublishAction` (which constructs its own `SdkPublishPrompts`, `PublishingApiService`, and `GenerateAction`) — either is larger and riskier than this parity fix. Verified manually instead.

Automated: `pnpm build` and `pnpm lint` clean.

Manual (against a real publishing profile):

1. Interactive, Package + Source Code → polls to `[Published] | [Published]`, exit `0`.
2. Interactive, package-only → one target in the status line.
3. Interactive, a publish that fails → exit `1`, log URL printed.
4. Interactive, Ctrl+C mid-poll → immediate exit, "still running" notice, log URL.
5. Non-interactive regression → output unchanged.


src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+      },
+      cancelMessage: 'Publishing is still running on APIMatic and will continue without the CLI.'
+    });
+
+    spin.start('Waiting for publishing status...');
+
+    while (!spin.isCancelled) {
+      const publishingLogResult = await getSdkPublishingLogFn();
+
+      if (publishingLogResult.isErr()) {
```
saeedjamshaid, line 71: **A single transient status-GET error is reported as publish failure.**

There is no retry tolerance here, and `getSdkPublishingLog` has no retry layer of its own — `publishing-api-service.ts:142` creates a bare `axios.create` with no `axios-retry`. One 502/503, a network blip, or a token expiring during a 15-minute Java publish makes this return `err`, and the loop immediately stops with `'failed'` → `ActionResult.failed()` → exit 1, while publishing continues and succeeds server-side.

Before this PR interactive mode always exited 0 after the POST, so this newly exposes interactive users to false failures; in CI it can trigger rollback or alerting on a healthy publish.

Suggest tolerating N consecutive fetch errors before giving up (or adding `axios-retry` to the publishing client).

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+
+      spin.message(statusMessage);
+      await new Promise<void>((resolve) => {
+        const timer = setTimeout(resolve, POLL_INTERVAL_MS);
+        abortWait = () => {
+          clearTimeout(timer);
+          resolve();
+        };
+      });
+      abortWait = undefined;
```
saeedjamshaid, line 106: **Ctrl+C landing while the status request is in flight is ignored for up to 10s.**

`abortWait` is assigned inside the Promise executor on line 101 and reset to `undefined` here. During `await getSdkPublishingLogFn()` (line 69) it is `undefined`, so the `onCancel` handler on line 61 (`abortWait?.()`) is a no-op. The user presses Ctrl+C mid-request: clack prints the cancel message and stops the spinner, then this code falls through to `spin.message(...)` on a dead spinner and blocks on the full `setTimeout(resolve, 10000)` before `while (!spin.isCancelled)` is re-evaluated. The terminal sits silent for up to 10 seconds after the cancel message.

Manual test 4 in the PR description ("Ctrl+C mid-poll → immediate exit") only holds when the interrupt happens to land inside the sleep window.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+            Succeeded: 'Published'
+          };
+          const label = eventLabels[event.eventType] ?? 'Failed';
+          return `${target}: [${label}]`;
+        })
+        .join(' | ');
+
+      if (executionCompleted) {
+        const isExecutionSuccessful = events.every((event) => event.eventType === 'Succeeded');
+        spin.stop(statusMessage, isExecutionSuccessful ? 0 : 1);
```
saeedjamshaid, line 94: **A post-cancel iteration returns `succeeded`/`failed` instead of `cancelled`, and stops the spinner a second time.**

The loop body never re-checks `spin.isCancelled` after the await. If Ctrl+C arrives while the fetch is in flight, clack has already stopped the spinner with the cancel message; the resolved fetch then reaches this line (or line 72) and calls `spin.stop(...)` again.

clack's `stop` has no already-stopped guard: it runs `clearPrevMessage()` — emitting cursor-up/erase over the line clack just printed — then writes a second symbol+message line and runs `clearHooks()`/`unblock()` twice. The user sees the cancel notice overwritten by `Source Code: [Published] | Package: [Published]`, and the function returns `'succeeded'` → exit 0 (or `'failed'` → exit 1) when the user's intent was a cancel (exit 130).

A `if (spin.isCancelled) return 'cancelled';` immediately after the await closes both halves.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+    while (!spin.isCancelled) {
+      const publishingLogResult = await getSdkPublishingLogFn();
+
+      if (publishingLogResult.isErr()) {
+        spin.stop('Failed to fetch publishing status.', 1);
+        return 'failed';
+      }
+
+      const { events } = publishingLogResult.value;
+      const executionCompleted = events.every((event) => TERMINAL_STATES.has(event.eventType));
```
saeedjamshaid, line 77: **An empty `events` array reports instant success.**

`events.every(...)` is vacuously `true` on `[]`. If the first `getSdkPublishingLog` returns `{ publishLogId, events: [] }` — the log row created before the per-target events are written, a normal race right after the POST — then `executionCompleted` is true, `isExecutionSuccessful` (line 93) is also true, `spin.stop('', 0)` prints a bare green check with no text, and `sdk publish` exits 0 within a second claiming a publish that has not started.

The PR defers this to #312, but the move promotes it from a non-interactive-only defect to the interactive default — interactive previously never asserted success or failure at all. A `events.length > 0 &&` guard is cheap insurance in the meantime.

src/actions/sdk/publish.ts
```diff
@@ -100,5 +100,38 @@ export class SdkPublishAction {
...
+    }
+
+    // Since publishing was not initiated, skip the next steps for polling.
+    if (dryRun) {
+      return ActionResult.success();
+    }
+
+    const publishingInfo = publishResult.getValue();
+    this.prompts.publishingRunningNotice(publishingProfile, language, semVersion, publishType);
+    this.prompts.publishingLogsMessage(publishingInfo.publishingLogUrl);
```
saeedjamshaid, line 118: **Non-interactive output did change**, contrary to "Non-interactive behaviour is unchanged" in the PR description.

Old order: runningNotice → poll → `postPublishingMessage` (title `Next Steps`, body "To view publishing logs, please visit:").
New order: runningNotice → `publishingLogsMessage` (title `Publishing Logs`, body "To track progress and view publishing logs, please visit:") → poll.

On a failed publish in CI the URL is no longer the last thing printed — it is buried above the failure line and above however many spinner frames the poll emitted, which is exactly the case the PR body says the URL is "essential" for. Any log scraping keyed on the `Next Steps` heading also breaks.

The PR description itself states the order as "running notice → poll → closing note with the log URL", which is not what this does. Either restore the trailing note or update the description.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+    const spin = spinner({
+      onCancel: () => {
+        abortWait?.();
+      },
+      cancelMessage: 'Publishing is still running on APIMatic and will continue without the CLI.'
+    });
+
+    spin.start('Waiting for publishing status...');
+
+    while (!spin.isCancelled) {
```
saeedjamshaid, line 68: **Unbounded poll with no stall detection — interactive `sdk publish` can now block forever.**

If one target stalls in `Queued` (misconfigured registry credentials, wedged backend worker), `TERMINAL_STATES` never matches, `executionCompleted` stays false, and this loop polls every 10s indefinitely with an unchanging status line. Interactive users, who before this PR got the log URL and an exit within seconds, now have a session pinned until they discover Ctrl+C.

A fixed timeout is admittedly a guess, but what ships here is an unbounded hang with no signal at all — even a "no state change in N minutes" warning would be an improvement.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+    );
+  }
+
+  public publishingLogsMessage(publishingLogUrl: string) {
+    const message = `To track progress and view publishing logs, please visit:
+${f.link(publishingLogUrl)}`;
+    noteWrapped(message, 'Publishing Logs');
+  }
+
+  public async pollPublishingStatus(
```
saeedjamshaid, line 52: **Architecture: the exit-code decision now lives in the Prompts layer.**

`.ai/instructions.md:49` — Prompts are "All terminal UI via `@clack/prompts` … **No business logic**"; line 47 assigns use-case orchestration and `ActionResult` to Actions.

`pollPublishingStatus` owns `TERMINAL_STATES`, `POLL_INTERVAL_MS`, the completion predicate, and `events.every(e => e.eventType === 'Succeeded')` — the value the action mechanically maps to the process exit code at `actions/sdk/publish.ts:128-135`. This PR promotes it from a non-interactive-only helper to the single shared decision point for both modes, so the violation now governs every `sdk publish` invocation.

`portal-service.ts:319 pollUntilCompleted` is the in-repo precedent for where a poll loop belongs. Moving it there would also give the status fetch a natural place to gain retry tolerance (see the comment on line 71).

src/actions/sdk/publish.ts
```diff
@@ -100,5 +100,38 @@ export class SdkPublishAction {
...
+      return ActionResult.cancelled();
+    }
+
+    // Since publishing was not initiated, skip the next steps for polling.
+    if (dryRun) {
+      return ActionResult.success();
+    }
+
+    const publishingInfo = publishResult.getValue();
+    this.prompts.publishingRunningNotice(publishingProfile, language, semVersion, publishType);
```
saeedjamshaid, line 117: **Interactive mode now prints the same Profile/Language/Version/Targets block twice, seconds apart.**

`SdkPublishInteractiveAction.execute` already calls `this.prompts.publishingSummary(...)` at `actions/sdk/publish/interactive.ts:89` ("Ready to publish:" + the four-field block); the user confirms, and then this shared path calls `publishingRunningNotice` with the identical four values ("Publishing is running for the following:" + the same block). Only the header differs.

Either skip `publishingRunningNotice` on the interactive path or fold it into the existing summary.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+  }
+
+  public publishingRunningNotice(
+    profile: PublishingProfile,
+    language: Language,
+    version: SemVersion,
+    publishType: PublishType[]
+  ): void {
+    const targets = [...publishType]
+      .sort((a, b) => (a === PublishType.SourceCodePublishing ? -1 : b === PublishType.SourceCodePublishing ? 1 : 0))
```
saeedjamshaid, line 38: **Not a valid sort comparator, and now triplicated across three call sites.**

`(a, b) => (a === SourceCodePublishing ? -1 : b === SourceCodePublishing ? 1 : 0)` returns `-1` when *both* operands are SourceCode — asserting `a < b` for equal elements, a violation of antisymmetry. It stays harmless today only because these arrays hold at most two distinct values.

The identical expression appears here, again at line 79 (keyed on `event.publishType === 'SourceCode'`), and at `prompts/sdk/publish/non-interactive.ts:70`. Any future publish type — or an events array carrying per-target history rather than one current event each — produces an arbitrary display order.

One shared rank function (`sourceCodeFirst`) used by all three sites fixes both problems.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+export class SdkPublishPrompts {
+  public publishSdk(fn: Promise<Result<PublishingInfo, ServiceError>>) {
+    return withSpinner('Publishing SDK', 'Publishing initiated.', 'SDK Publishing failed.', fn);
+  }
+
+  public sdkPublishingServiceError(serviceError: ServiceError) {
+    log.error(serviceError.errorMessage);
+  }
+
+  public dryRunNotice(publishingProfile: PublishingProfile, language: Language, version: SemVersion, publishType: PublishType[]): void {
```
saeedjamshaid, line 24: **`dryRunNotice` and `publishingRunningNotice` are copy-paste variants that disagree on target ordering.**

The two methods sit 7 lines apart and build the same Profile/Language/Version/Targets layout from the same four parameters. `dryRunNotice` (line 25) maps without sorting; `publishingRunningNotice` (line 37) sorts source-code first. For the same `publishType` input, `--dry-run` can print `Package + Source Code` while the real publish prints `Source Code + Package`, so a user comparing a dry run against the real run sees a spurious difference.

Extract one `formatTargets(publishType)` and one block builder used by both — and by `publishingSummary` in the interactive prompts, which is a third copy.

src/actions/sdk/publish.ts
```diff
@@ -100,5 +100,38 @@ export class SdkPublishAction {
...
+
+    if (publishResult.isFailed()) {
+      return ActionResult.failed();
+    }
+    if (publishResult.isCancelled()) {
+      return ActionResult.cancelled();
+    }
+
+    // Since publishing was not initiated, skip the next steps for polling.
+    if (dryRun) {
```
saeedjamshaid, line 112: **The `dryRun` early-return is duplicated inside and outside the `withDirPath` block for a single condition.**

`if (dryRun)` at line 73 returns `ActionResult.success()` with no payload from a callback annotated `Promise<ActionResult<PublishingInfo>>`. This outer guard exists solely to intercept that valueless success before `getValue()` at line 116 returns `undefined` and the subsequent property reads crash.

Two branches encode one decision, and the type-level hazard remains latent — deleting this guard silently produces `publishingInfo.publishLogUrl` on `undefined` with no compiler complaint. Splitting the dry-run path out of `execute`, or returning a discriminated result from the inner block, removes both the duplication and the hazard.

src/actions/sdk/publish.ts
```diff
@@ -100,5 +100,38 @@ export class SdkPublishAction {
...
+        this.commandMetadata.shell
+      )
+    );
+
+    switch (publishingOutcome) {
+      case 'succeeded':
+        return ActionResult.success();
+      case 'cancelled':
+        return ActionResult.cancelled();
+      default:
```
saeedjamshaid, line 133: **`default:` instead of `case 'failed':` defeats the point of the `PublishingOutcome` union.**

The union was introduced deliberately because "a boolean cannot express cancel-vs-fail, and the distinction matters for planned work". Adding `'timed_out'` — the exact planned work the PR names — compiles clean and falls into `default`, returning `ActionResult.failed()` and exit 1 for a timeout, with no build-time signal that this site needs updating.

```ts
case 'failed':
  return ActionResult.failed();
default: {
  const _never: never = publishingOutcome;
  return _never;
}
```

saeedjamshaid (commented):
Automated review of the polling change. 15 findings: 6 correctness issues in the new `pollPublishingStatus` loop and the action wiring, 1 architecture note, and 8 quality/hygiene items. Ordered roughly most-severe first in the inline comments below.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+    noteWrapped(message, 'Publishing Logs');
+  }
+
+  public async pollPublishingStatus(
+    getSdkPublishingLogFn: () => Promise<Result<PublishLogItem, ServiceError>>
+  ): Promise<PublishingOutcome> {
+    const TERMINAL_STATES = new Set(['Succeeded', 'Failed', 'Exception', 'InternalError']);
+    const POLL_INTERVAL_MS = 10000; // poll after every 10 seconds.
+
+    let abortWait: (() => void) | undefined;
```
saeedjamshaid, line 58: **The mutable `abortWait` latch hand-rolls a cancellable sleep; an `AbortController` is simpler and closes the race above.**

`abortWait` is declared here, assigned inside a Promise executor at line 101, and reset at line 106 — a three-site mutable-latch dance whose correctness depends entirely on that assignment window, which is precisely why the cancel is missed during the fetch.

Hoisting an `AbortController` that `onCancel` fires, then `await sleep(POLL_INTERVAL_MS, controller.signal)` plus a `signal.aborted` check after the fetch, is fewer moving parts and makes the whole iteration cancellable rather than just the sleep. clack's `SpinnerOptions` already accepts a `signal` (`index.d.mts:246`), so the controller has a second use.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
-  }
-
-  public dryRunNotice(publishingProfile: PublishingProfile, language: Language, version: SemVersion, publishType: PublishType[]): void {
-    const targets = publishType.map((t) => (t === PublishType.PackagePublishing ? 'Package' : 'Source Code')).join(' + ');
-    log.info(
-      `You can publish this SDK by removing the --dry-run flag. It will be published for the following:\n\n  Profile:   ${publishingProfile}\n  Language:  ${language}\n  Version:   ${version}\n  Targets:   ${targets}`
-    );
-  }
-}
+import { log, spinner } from '@clack/prompts';
```
saeedjamshaid, line 1: **This file's line endings flipped from CRLF to LF, so the entire file reads as changed and blame is destroyed.**

`git cat-file blob origin/dev:src/prompts/sdk/publish.ts` is CRLF; the same blob at HEAD is LF. `git diff origin/dev...HEAD` therefore shows all 137 lines removed and re-added, while `--ignore-all-space` shows the true change is 87 insertions / 2 deletions. A reviewer cannot see what actually changed, and `git blame` now attributes `publishSdk`, `sdkPublishingServiceError`, and `dryRunNotice` to this commit.

The repo has no `.gitattributes` and local `core.autocrlf=true`, so mixed-EOL blobs will keep producing these phantom rewrites. Adding `* text=auto eol=lf` would settle it repo-wide.

src/prompts/sdk/publish.ts
```diff
@@ -1,26 +1,111 @@
...
+    const message = `To track progress and view publishing logs, please visit:
+${f.link(publishingLogUrl)}`;
+    noteWrapped(message, 'Publishing Logs');
+  }
+
+  public async pollPublishingStatus(
+    getSdkPublishingLogFn: () => Promise<Result<PublishLogItem, ServiceError>>
+  ): Promise<PublishingOutcome> {
+    const TERMINAL_STATES = new Set(['Succeeded', 'Failed', 'Exception', 'InternalError']);
+    const POLL_INTERVAL_MS = 10000; // poll after every 10 seconds.
```
saeedjamshaid, line 56: `.ai/instructions.md:70` — "**No AI-generated comments** — comment only to explain non-obvious *why*: a constraint, a workaround, a subtle invariant. Never restate what the code already says."

`// poll after every 10 seconds.` says nothing the identifier and the literal do not. The interval is also the one number here that would benefit from a real *why* (why 10s — what backend cadence does it track?), so the comment occupies the slot without carrying the information.

saeedjamshaid (approved):

PR #318, opened 2026-08-13
feat: create and maintain plugin-config.json from sdk publish and plugin generate

MuHamza30:
Follows #314, which added `apimatic plugin generate` for the happy path only — it assumed `src/plugin-config.json` already existed and named at least one language. This adds everything that creates and maintains that file.

## The split

**`sdk publish` records languages. `plugin generate` owns metadata.** Publishing never asks for a plugin ID and never calls the account API, so a project can be published from long before anyone decides to build a context plugin.

- **`sdk publish`** — after a successful publish, records the published language. Interactive asks; non-interactive answers with `--update-plugin-config`, because a prompt on the CI path would never be answered.
- **`plugin generate`** — prompts for plugin ID / name / version when they are missing, then generates; or stops with next steps when no language has been recorded yet.

## What gets written

```jsonc
{
  "schemaVersion": 1,
  "pluginId": "acme-payments",          // metadata: plugin generate only
  "pluginName": "Acme Payments",
  "pluginVersion": "0.1.0",
  "author": { "name": "Acme", "email": "developers@acme.com" },
  "license": "MIT",                     // written unprompted; the backend consumes it
  "languages": {                        // sdk publish only
    "csharp": {
      "source":  { "repositoryUrl": "https://github.com/acme/acme-payments-csharp", "branch": "main" },
      "package": { "packageId": "Acme.Payments.Sdk" },
      "version": "v3"                   // the codegen version the publish actually used
    }
  }
}
```

Package shape is per language, mapped from the publishing profile: `csharp: packageId` · `typescript|python|ruby: name` · `java: groupId+artifactId` · `php: vendorName+projectName` · `go: packageName`.

`pluginKey` is deliberately never written — codegen-v2 parses and validates it, but `SdkReposExtensions.ResolveAgainst` never reads it. A read-modify-write round-trip preserves fields this CLI version does not model, and backfills `schemaVersion` when a hand-written file omits it.

## `plugin generate` states

| Config | Behaviour |
|---|---|
| unusable (unparseable, or a `schemaVersion` this CLI does not model) | error naming the file and reason, `failed()` |
| missing / no metadata | prompt for metadata, then the language check below |
| no languages | next steps, `success()`, no generation |
| metadata + languages | generate |

## An entry only ever claims what was published

Three cases where the CLI deliberately records nothing rather than record something untrue — each says so rather than exiting quietly:

- **A dry run.** `--dry-run` publishes nothing but returns success, so `--update-plugin-config` is ignored with a notice.
- **A half not published.** The entry is built from the run's `--publish-type`, not from whatever the profile enables — so a package-only run records no `source`, and a source-only run records no `package`.
- **A repository name that cannot be resolved.** A publishing profile carries no git host (`GitConfigurationItem` has only `credentialsId`, `repositoryName`, `branch`), so only an absolute `http(s)` URL or a plain `owner/repo` can be turned into a URL. Anything else — an SSH remote, a bare single-segment name — is refused and reported with the offending value, rather than guessed at. **This is the design decision most worth a second opinion**: the backend validates only that `repositoryUrl` is non-blank, so a wrong guess would survive to a failed clone, but the guard also means a profile shape we have not seen in the wild records nothing.

## Notes for review

- **`--update-plugin-config` is non-interactive only.** `sdk publish` treats *any* argument as non-interactive (`this.argv.length === 0`), so the flag cannot be combined with the interactive flow; interactive asks instead.
- **Recording can never fail a publish.** A write fault is reported, not thrown — the publish that already pushed the package must not become a stack trace because a config file was locked or the disk was full.
- **`license: "MIT"` is written unprompted**, and re-added if a user removes it. Flagging rather than deciding.
- **A run with no languages exits 0** having generated nothing. Also flagging rather than deciding — `failed()` would make the no-op detectable by a script.

## Known follow-ups

- README is not regenerated — `--update-plugin-config` is undocumented. `pnpm readme` duplicates every section on a CRLF checkout, so this needs care.
- The read side does not verify that an *existing* language entry carries a `source`; a malformed hand-written config is caught by the backend rather than locally. `isLanguageEntry` was written for that check and is currently unused.
- Layering: the `publishTypes` gating sits in `PluginRecordSdkAction` while the rest of the entry's shape is decided in `buildLanguageEntry` — one rule in two places. And `buildLanguageEntry` is a pure transformation living in `src/types/` rather than `src/application/`.
- `PluginRecordSdkAction.execute` returns `void` rather than `ActionResult`, so that recording cannot alter the publish result. Callers ignoring the result would be the better shape.
- The publish → record wiring has no test seam: actions declare `execute` as an arrow-function field, so it cannot be stubbed on the prototype.
- `plugin-config.json` appears as a literal in several prompts classes rather than one shared constant.

## Testing

`pnpm build` and `pnpm lint` are clean. 194 passing, 11 pending — excluding `test/prompts/sdk/generate.test.ts` and `test/prompts/sdk/publish/publishing-details.test.ts`, which call `CodegenOption.resolve` and abort the suite. **That breakage is pre-existing on `dev` from #317** — the method is named `create` — and is not introduced here.


src/prompts/plugin/generate.ts
```diff
@@ -43,6 +46,30 @@ export class PluginGeneratePrompts {
...
+  }
+
+  /** Reached on a successful run that found nothing to build, not on a generation failure. */
+  public nextStepsPublishSdks() {
+    const message =
+      `Publish an SDK for each language you want in the plugin:\n` +
+      `'${f.cmdAlt('apimatic', 'sdk', 'publish')} ${f.flag('language', '<language>')}'\n` +
+      `Then run '${f.cmdAlt('apimatic', 'plugin', 'generate')}'.`;
+    noteWrapped(message, 'Next Steps');
+  }
```
mrafnadeem-apimatic, line 70: This should now recommend using the interactive publishing command since that would be easier for the user. Also, it should tell the user that "Source Code" publishing is required.

src/types/plugin/plugin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+  | CSharpPluginPackage
+  | NamedPluginPackage
+  | JavaPluginPackage
+  | PhpPluginPackage
+  | GoPluginPackage;
+
+export interface LanguageEntry {
+  source: LanguageSource;
+  package?: PluginPackage;
+  version?: CodeGenerationVersion;
```
mrafnadeem-apimatic: why is this optional? it should be required.

src/types/plugin/plugin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export interface PluginConfigData {
+  schemaVersion: number;
+  // Optional on disk: `sdk publish` creates a config carrying languages alone, and
+  // `plugin generate` fills these in before it ever uploads.
+  pluginId?: string;
+  pluginName?: string;
+  pluginVersion?: string;
+  // Read and preserved on a round-trip, never written: nothing in codegen-v2 consumes it.
+  pluginKey?: string;
```
mrafnadeem-apimatic: It should be required and populated because we will eventually use this for telemetry.

src/types/plugin/plugin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+  | GoPluginPackage;
+
+export interface LanguageEntry {
+  source: LanguageSource;
+  package?: PluginPackage;
+  version?: CodeGenerationVersion;
+}
+
+/** The backend also accepts a bare source URL in place of an entry. Read, never written. */
+export type LanguageValue = LanguageEntry | string;
```
mrafnadeem-apimatic: String case is not supported (and is difficult to implement in C# anyways). Remove the union and keep the LanguageEntry only.

src/types/plugin/plugin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+export interface PhpPluginPackage {
+  vendorName: string;
+  projectName: string;
+  packageUrl?: string;
+}
+
+export interface GoPluginPackage {
+  packageName: string;
+  packageUrl?: string;
+}
```
mrafnadeem-apimatic: `packageUrl` is never populated so it shouldn't be added at all.

src/prompts/plugin/record-sdk.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Language } from '../../types/sdk/generate.js';
+import { format as f } from '../format.js';
+
+const PLUGIN_CONFIG_FILE = 'plugin-config.json';
+
+export class PluginRecordSdkPrompts {
+  public async confirmRecordSdk(language: Language, configExists: boolean): Promise<boolean> {
+    const message = configExists
+      ? `Add ${f.var(language)} to ${f.var(PLUGIN_CONFIG_FILE)}?`
+      : `Create ${f.var(PLUGIN_CONFIG_FILE)} and add ${f.var(language)} to it?`;
```
mrafnadeem-apimatic: This doesn't give the user any context on what context plugins are if they are seeing this prompt for the first time. A link to a showcase, or demo would be valuable here.

src/types/plugin-config-context.ts
```diff
@@ -0,0 +1,147 @@
...
+export class PluginConfigContext {
+  private readonly fileService = new FileService();
+
+  constructor(private readonly buildDirectory: DirectoryPath) {}
+
+  private get configPath(): FilePath {
+    return new FilePath(this.buildDirectory, new FileName('plugin-config.json'));
+  }
+
+  public async validate(): Promise<PluginConfigState> {
```
saeedjamshaid: this method should be renamed to something like getPluginConfigState

src/types/plugin-config-context.ts
```diff
@@ -0,0 +1,147 @@
...
+  PluginMetadata
+} from './plugin/plugin-config.js';
+import { Language } from './sdk/generate.js';
+
+/**
+ * What a caller needs to know before generating. Metadata and languages are written by different
+ * commands — `plugin generate` owns the first, `sdk publish` the second — so they are reported
+ * separately. `path` rides on `unreadable` purely so the prompt can say where to fix the file.
+ */
+export type PluginConfigState =
```
saeedjamshaid, line 21: Convert this type into a class with private fields. And add methods like HasLanguages(), HasMetadata() etc. so that all of the logical operations and state checks are encapsulated inside these methods

saeedjamshaid (changes requested):

mrafnadeem-apimatic (changes requested):

src/utils/string-utils.ts
```diff
@@ -9,6 +9,21 @@ export const removeQuotes = (input: string): string => {
...
 
+/** Lower-case kebab-case — the shape a plugin id must take to pass server-side validation. */
+export const toKebabCase = (input: string): string =>
+  input
+    .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '-')
+    .replace(/^-+|-+$/g, '');
+
+export const toTitleCase = (input: string): string =>
```
mrafnadeem-apimatic: A one-line comment telling us which style of title case is used here would be good.

src/utils/string-utils.ts
```diff
@@ -9,6 +9,21 @@ export const removeQuotes = (input: string): string => {
...
+  input
+    .split(/[^a-zA-Z0-9]+/)
+    .filter(Boolean)
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
```
mrafnadeem-apimatic: These kinds of functions are good candidates for unit tests.

src/utils/string-utils.ts
```diff
@@ -9,6 +9,21 @@ export const removeQuotes = (input: string): string => {
   return input;
 };
 
+/** Lower-case kebab-case — the shape a plugin id must take to pass server-side validation. */
+export const toKebabCase = (input: string): string =>
```
mrafnadeem-apimatic: Try out a more robust solution like:
```ts
export const toKebabCase = (input: string) => 
  input && 
  input
    .match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)
    .map(x => x.toLowerCase())
    .join('-');
```

Unit tests are mandatory for this function.

src/prompts/plugin/create-config.ts
```diff
@@ -0,0 +1,81 @@
...
+   * Required fields get a `placeholder` but no `defaultValue`, so an empty answer re-prompts.
+   * Optional ones get both, so Enter accepts the suggestion — the same split `sdk quickstart` uses.
+   */
+  public async inputPluginMetadata(defaults: PluginMetadata): Promise<PluginMetadata | undefined> {
+    const pluginId = await text({
+      message: 'Enter an ID for your plugin:',
+      placeholder: defaults.pluginId,
+      validate: (value) => {
+        if (!value) return 'Plugin ID is required.';
+        if (!KEBAB_CASE.test(value)) return `Plugin ID must be lower-case kebab-case, for example 'acme-payments'.`;
```
mrafnadeem-apimatic: also mention that it should be alphanumeric

src/prompts/plugin/generate.ts
```diff
@@ -43,6 +46,30 @@ export class PluginGeneratePrompts {
...
 
+  public pluginConfigUnreadable(reason: string, path: FilePath) {
+    const message =
+      `${f.var(PLUGIN_CONFIG_FILE)} cannot be used: ${reason}. ` + `Fix or delete it at ${f.path(path)} and try again.`;
+    log.error(message);
+  }
+
+  /** A cancelled metadata prompt stops the run, so it is a warning rather than an error. */
+  public metadataCancelled() {
+    log.warn('A plugin ID is required. Exiting without generating a plugin.');
```
mrafnadeem-apimatic: This appears even if the user has already entered the plugin ID. It should instead accept a parameter named "reason" which it can print. The reason should come from the action that tries creating a `plugin-config.json`.

mrafnadeem-apimatic (changes requested):

src/actions/plugin/record-sdk.ts
```diff
@@ -0,0 +1,81 @@
...
+      codegenVersion
+    );
+
+    // Both of these return before the confirm below, so without a message the prompt would simply
+    // never appear.
+    if (built.kind === 'unresolvableRepositoryName') {
+      this.prompts.unresolvableRepositoryName(language, built.repositoryName);
+      return;
+    }
+    if (built.kind !== 'entry') {
```
saeedjamshaid: In case source code publishing didn't happen, we should still add the published package details in the plugin-config.json and just warn the user that plugin won't be generated without source code details

src/actions/plugin/record-sdk.ts
```diff
@@ -0,0 +1,81 @@
...
+        : undefined,
+      publishTypes.includes(PublishType.PackagePublishing)
+        ? publishingProfile.getPackageConfigurationDataForLanguage(language)
+        : undefined,
+      codegenVersion
+    );
+
+    // Both of these return before the confirm below, so without a message the prompt would simply
+    // never appear.
+    if (built.kind === 'unresolvableRepositoryName') {
```
saeedjamshaid: Impossible state since the publishing profile will always have valid github repo url

src/types/plugin-config-context.ts
mrafnadeem-apimatic: renamed to `loadState`

src/prompts/plugin/generate.ts
mrafnadeem-apimatic: Done in 287897e9. `nextStepsPublishSdks` now points at the interactive `apimatic sdk publish` (no `--language`) and says the SDK must include `Source Code` publishing.

src/types/plugin/plugin-config.ts
mrafnadeem-apimatic: Done. The codegen version is required on every language entry now.

src/types/plugin/plugin-config.ts
mrafnadeem-apimatic: Leaving this out for now — it is scope creep for the bootstrap work. `pluginKey` stays optional and unwritten: f16a0e84 populated it from the account's API Copilot key, and cd78cac9 pulled that back out. We will pick it up when telemetry actually needs it.

src/types/plugin/plugin-config.ts
mrafnadeem-apimatic: Done. `LanguageValue` and the string arm are gone; `PluginLanguages` maps a language straight to its entry.

src/types/plugin/plugin-config.ts
mrafnadeem-apimatic: Done in 4c284a04. `packageUrl` was dropped from every package type, along with the other fields nothing writes.

src/prompts/plugin/record-sdk.ts
mrafnadeem-apimatic: For now the first-run prompt points at `apimatic plugin generate --help`, which says what a context plugin is and does (7290d6b1). A showcase or demo link needs documentation and marketing material that does not exist yet — we will link it once that is published.

src/types/plugin-config-context.ts
mrafnadeem-apimatic: Done in 8a1a269a. The `present` case is a `PluginConfigPresent` class holding the config in a private field, with `hasPublishedSdks()` and `hasMetadata()` on it, so callers never touch the raw data. `missing` and `unreadable` stay plain variants — they carry no config to encapsulate, only the reason and path the prompt needs.

src/utils/string-utils.ts
mrafnadeem-apimatic: Moot — `toTitleCase` was removed in 46ef30a when the derived defaults were replaced with fixed example values.

src/utils/string-utils.ts
mrafnadeem-apimatic: Covered then removed with the functions themselves (46ef30a); nothing left to test here.

src/utils/string-utils.ts
mrafnadeem-apimatic: Both helpers were hardened and unit-tested in 93df418/aa93db5, then removed entirely in 46ef30a: the prompts no longer derive a plugin ID or name from anything, they just suggest a fixed example (`acme-payments` / `Acme Payments`). No case conversion left in the codebase.

src/prompts/plugin/create-config.ts
mrafnadeem-apimatic: Done in fae64416. The message now reads: "Plugin ID must be lower-case alphanumeric words separated by single dashes, for example 'acme-payments'."

src/prompts/plugin/generate.ts
mrafnadeem-apimatic: Done in 428c2d38. `metadataCancelled(reason)` takes the reason, and it comes from the create-config action: the metadata prompt returns which answer was missing, so cancelling on the name says "A plugin name is required".

saeedjamshaid (approved):

mrafnadeem-apimatic (approved):

src/actions/plugin/record-sdk.ts
mrafnadeem-apimatic: done

src/actions/plugin/record-sdk.ts
mrafnadeem-apimatic: removed this case

PR #321, opened 2026-08-20
feat: add apimatic plugin publish

### LINKED ISSUE #320: Plugin publish command
# Add `apimatic plugin publish` — print the git commands for publishing a context plugin to GitHub

## Description

Adds a command that **prints** the git commands needed to publish a generated context plugin to GitHub. The CLI never runs them — no `git`, no `gh`, no network call, no authentication.

The command reads the plugin directory and `src/plugin-config.json`, then prints one of two command blocks depending on whether the plugin directory is already a git repository. It writes nothing.

## Output — first publish (no `plugin/.git`)

```
Commands to Copy and Run ──────────────────────────────────────────

Publish version '0.1.67' of your plugin to a new public GitHub repository.

What is about to become public: 214 files, 4 dirs

cd "C:\Users\Hamza\Desktop\acme\plugin"
git init -b main
git add .
git commit -m "Publish hamza 0.1.67"
git tag -a v0.1.67 -m "hamza 0.1.67"
gh repo create hamza --public --source=. --remote=origin
git push -u origin main --follow-tags
```

## Output — update (`plugin/.git` present)

```
Commands to Copy and Run ──────────────────────────────────────────

Publish version '0.1.67' of your plugin as a new release of your existing repository.

What is about to become public: 214 files, 4 dirs

cd "C:\Users\Hamza\Desktop\acme\plugin"
git add .
git commit -m "Publish hamza 0.1.67"
git tag -a v0.1.67 -m "hamza 0.1.67"
git push --follow-tags
```

`hamza` is `pluginId`; `0.1.67` and `v0.1.67` come from `pluginVersion`. Nothing else is read from the config.

## Why these commands

- **One command per line.** No `&&` chaining, so a user can paste them one at a time and see which step failed.
- **`git add .` before committing, not `git commit -am`.** `-a` stages only files git already tracks. On a first publish nothing is tracked yet, so `-am` commits nothing and exits 1. On a re-release it silently omits any file the regenerated plugin added.
- **`git tag -a`, not `git tag`.** `git push --follow-tags` carries annotated tags only; a lightweight tag would stay on the user's machine while the commit reached GitHub, so the release would never appear.
- **Both blocks share the same `add → commit → tag` sequence**, so the two flows cannot drift apart.

## Why two blocks

Updates are the dominant case — a first publish happens once per plugin, releases happen every version. Printing the first-publish block on an update gives the user a hard failure (`gh repo create` → "Name already exists on this account") and a noisy no-op (`git init` → "Reinitialized existing Git repository").

## Flags

- `-i, --input` — parent directory containing `src`
- `-d, --destination` — the plugin directory, described as where the plugin **was** generated (publish reads an artifact that already exists)

No `--force` (nothing is written) and no `--auth-key` (no API call).

## Validations

In order. All must pass before anything is printed. Every failure exits 1 and names the next step.

| # | Check | Message points to |
|---|---|---|
| 1 | Plugin directory is not the build directory | the conflicting path |
| 2 | Plugin directory has content | `apimatic plugin generate` |
| 3 | `plugin-config.json` is readable and its identity fields are well-formed | the file path, then `apimatic plugin generate` |
| 4 | `plugin-config.json` exists | `apimatic plugin generate` |
| 5 | `pluginId` and `pluginVersion` are both recorded | `apimatic plugin generate` |

**#1** matters more here than in `generate`: the printed block ends in a `--public` repository containing `git add .` of that directory. If it points at `src/`, the user publishes their API specs publicly.

**#3 vs #5** — a field that is present but malformed (a bad id, a non-semver version, a blank or padded value) means the file was hand-edited, so the file is refused and the reason names the field. A field that is simply absent is normal: `sdk publish --update-plugin-config` writes a config carrying languages alone, which `plugin generate` names afterwards, so that case asks the user to run `plugin generate` instead.

## Deliberately not validated

All consequences of printing rather than executing:

- whether `git` or `gh` is installed, or `gh` is authenticated
- whether the directory is a repo, has commits, has `origin`, or its branch name
- whether the GitHub repo or the tag already exists
- whether `pluginVersion` was bumped since the last publish

## Known limitation, accepted

If a user deletes their local `plugin` folder and regenerates, no `.git` is found, so the first-publish block prints even though the GitHub repo exists — `gh repo create` then fails with "name already exists". Detecting this would require querying GitHub, turning a printing command into a networked one.

## Related change

`apimatic plugin generate` loses its `--zip` flag. A zipped plugin cannot be published: the printed commands would commit `plugin.zip` to a public repository, producing something nobody can install, and generate advertised that publish on its way out. Removing the flag leaves the plugin directory one shape, so the contents a publish reports are always the files it is about to expose. The flag never shipped.

## Acceptance criteria

- [ ] Prints the first-publish block when `plugin/.git` is absent, the update block when present
- [ ] Every validation fails with a distinct, actionable message and exit code 1
- [ ] The success path exits 0 and writes nothing to disk
- [ ] No network call, no authentication
- [ ] `cd` prints the resolved plugin directory in double quotes, so paths with spaces work on Windows
- [ ] **Every printed command succeeds when pasted, in both flows**
- [ ] Unit tests cover every validation branch and both output blocks
- [ ] `plugin publish` is registered in the examples-parse test and its examples parse
- [ ] README updated

MuHamza30:
closes #320 

Adds `apimatic plugin publish`, which **prints** the git commands for publishing a generated context plugin to GitHub. It never runs them — no `child_process`, no network call, no auth key.

Also fixes `plugin generate` deleting the user's `.git`, which this feature cannot work around.

## What it prints

`pluginId` and `pluginVersion` come from `src/plugin-config.json`. Nothing else is read.

**First publish** (no `plugin/.git`):

```
cd "C:\work\acme\plugin"
git init -b main
gh repo create hamza --public --source=. --remote=origin
git add . && git commit -m "Publish hamza 0.1.67"
git push -u origin main
git tag v0.1.67 && git push origin v0.1.67
```

**Update** (`plugin/.git` present):

```
cd "C:\work\acme\plugin"
git add . && git commit -m "Publish hamza 0.1.67"
git push
git tag v0.1.67 && git push origin v0.1.67
```

Both under a `printed, not executed` note, preceded by `what is about to become public: 4 files, 4 dirs`.

**Why two blocks:** a tag only matters when there is more than one version, so updates are the common case. The first-publish block on an update run fails at `gh repo create` ("Name already exists") and no-ops at `git init`.

**Four choices that are load-bearing:**

- `-b main` on `git init` — otherwise the branch comes from `init.defaultBranch` (still `master` on older installs) and `git push -u origin main` fails
- `gh repo create` without `--push` — no commits exist yet at that point
- `add`/`commit` before tagging — a tag points at a commit, so tagging uncommitted files publishes the *previous* snapshot silently
- `&&` on only the two inseparable pairs — the rest stay separate so a `gh` auth failure doesn't look like a git problem

## The `.git` fix

`PluginContext.save()` used `cleanDirectory` (`fsExtra.emptyDir`), which removed `.git` along with everything else. Regenerating a published plugin silently destroyed the local repository — history, `origin`, and all local tags — with no warning.

```
apimatic plugin generate
cd plugin && git init && git remote add origin <url> && git add . && git commit -m "initial"
apimatic plugin generate --force
cd plugin && git status     # fatal: not a git repository
```

Fixed with `cleanDirectoryExcluding(pluginDirectory, [new FileName('.git')])`, as `save-changes-context.ts:43` already does for SDK review directories.

Not separable from the feature: without it `.git` never survives a regeneration, so block selection always says "first publish".

## Validations

Six, in order; each fails with a distinct message and exit code 1.

| # | Check | Note |
|---|---|---|
| 1 | Plugin dir ≠ build dir | The block ends in a `--public` repo of `git add .`; pointed at `src` it publishes the user's API specs. `-d` resolves independently of `-i`, so nothing else stops `-d ./src` |
| 2 | Plugin dir has content | `directoryEmpty` filters dot-entries, so this also covers a directory holding only the preserved `.git` |
| 3 | `plugin-config.json` exists | |
| 4 | `plugin-config.json` parses | Reuses the existing reason text, so a broken file reads the same from both commands |
| 5 | `pluginId` present, kebab-case | Becomes the repository name |
| 6 | `pluginVersion` present, semver | Least protected field: optional on disk and `hasMetadata()` doesn't check it, yet it drives the version text, commit message and tag |

**Not validated**, all because we print rather than execute: whether `git`/`gh` exist or are authenticated; repo/commit/remote/branch state; whether the GitHub repo or tag already exists; whether the version was bumped. The `&&` before the tag push already stops a stale re-tag.

## Commits

| | |
|---|---|
| `style:` | Prettier on 4 drifted files — skippable |
| `fix:` | `.git` preservation + regression test |
| `feat:` | `PluginRelease`, `PluginContents`, `Directory.excluding/countFiles/countDirectories`, `getPluginRelease`, shared plugin-ID pattern |
| `feat:` | Command, action, prompts, tests, examples-parse, README, pointer from `plugin generate` |

Each functional commit builds and passes tests on its own. The `style:` commit exists because Prettier had drifted on these files (lint-staged only formats *staged* files) and folding it in would have buried the one-line `.git` fix under ~57 lines of reformatting.

## Design notes

- `PluginRelease` owns the `v` tag prefix and the plugin-ID pattern, each defined once; `record-metadata.ts` now imports the pattern instead of duplicating it
- `getPluginRelease()` is narrower than `hasMetadata()` — it needs the ID and version, never `pluginName`, so a config missing only the display name still publishes
- Counts exclude `.git`: `git add .` never stages it and its objects would dwarf the plugin
- The `cd` line uses double quotes, not `f.path()` — `f.path`'s single quotes aren't quoting to `cmd.exe`, so a path with spaces would break the pasted line. Only deviation from the format helpers, commented in place
- No `--force`, `--auth-key`, `CommandMetadata` or `getConfigDir()`; the action takes no constructor arguments


src/prompts/plugin/publish.ts
```diff
@@ -0,0 +1,110 @@
...
+  public pluginConfigMissing(directory: DirectoryPath) {
+    const message =
+      `${f.var(PLUGIN_CONFIG_FILE)} was not found in ${f.path(directory)}. ` +
+      `Run '${f.cmdAlt('apimatic', 'plugin', 'generate')}' first.`;
+    log.error(message);
+  }
+
+  public pluginConfigUnreadable(reason: string, path: FilePath) {
+    const message =
+      `${f.var(PLUGIN_CONFIG_FILE)} cannot be used: ${reason}. ` + `Fix or delete it at ${f.path(path)} and try again.`;
```
mrafnadeem-apimatic, line 60: should say "run apimatic plugin generate" instead of "try again"

src/types/plugin/plugin-release.ts
```diff
@@ -0,0 +1,58 @@
...
+      return err({ field: 'pluginVersion', reason: 'missing' });
+    }
+
+    const version = SemVersion.tryCreate(rawVersion);
+    if (version.isErr()) {
+      return err({ field: 'pluginVersion', reason: 'malformed' });
+    }
+
+    return ok(new PluginRelease(id, version.value));
+  }
```
mrafnadeem-apimatic: There is already validation in the parse method of PluginConfigContext. Move this there.

You don't need the getters in this class after you have already validated the object when you parsed it.

In that case, if it fails, it would be in an unreadable state. Of course, we should give the user good error messages for that case.

mrafnadeem-apimatic (changes requested):

src/types/plugin-context.ts
```diff
@@ -18,8 +21,22 @@ export class PluginContext {
     return !(await this.fileService.directoryEmpty(this.pluginDirectory));
   }
 
+  /** Counts exclude the repository: `git add .` never stages `.git`, and its objects would dwarf the plugin. */
+  public async describeContents(): Promise<PluginContents> {
+    const directory = await this.fileService.getDirectory(this.pluginDirectory);
+    const publishable = directory.excluding([new FileName(GIT_DIRECTORY_NAME)]);
+
+    return { fileCount: publishable.countFiles(), directoryCount: publishable.countDirectories() };
+  }
```
mrafnadeem-apimatic, line 25: Doesn't handle the case where plugin file is a zip.

src/actions/plugin/publish.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    const release = configState.getPluginRelease();
+    if (release.isErr()) {
+      this.prompts.pluginReleaseUnusable(release.error);
+      return ActionResult.failed();
+    }
+
+    const contents = await pluginContext.describeContents();
+
+    // A repository already here means every later release, where creating and linking it would fail.
```
mrafnadeem-apimatic: Look for and remove all redundant comments from this branch. If a comment is needed, the code must be changed to make the "why" obvious.

src/prompts/plugin/publish.ts
mrafnadeem-apimatic, line 1: Here's an example using a simpler set of commands(so that it looks like less work to the user):
```bash
cd "C:\Users\mrafn\Downloads\cli-plugin-qa\cli-plugin-qa\b2-v3-csharp\plugin"
git init -b main
gh repo create acme-payments --public --source=. --remote=origin --push
git commit -am "Publish acme-payments 1.0.0"
git tag v1.0.0
git push --follow-tags
```

src/prompts/plugin/publish.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (problem.field === 'pluginId') {
+      return `must be lower-case alphanumeric words separated by single dashes, for example ${f.var('acme-payments')}`;
+    }
+    return `must be a version in the format major.minor.patch, for example ${f.var('0.1.0')}`;
+  }
+
+  private contentsLine(contents: PluginContents): string {
+    const files = `${contents.fileCount} ${contents.fileCount === 1 ? 'file' : 'files'}`;
+    const directories = `${contents.directoryCount} ${contents.directoryCount === 1 ? 'dir' : 'dirs'}`;
+    return `what is about to become public: ${files}, ${directories}`;
```
mrafnadeem-apimatic:
```suggestion
    return `What is about to become public: ${files}, ${directories}`;
```

src/prompts/plugin/publish.ts
```diff
@@ -0,0 +1,110 @@
+import { log } from '@clack/prompts';
+import { DirectoryPath } from '../../types/file/directoryPath.js';
+import { FilePath } from '../../types/file/filePath.js';
+import { PluginContents } from '../../types/plugin/plugin-contents.js';
+import { PluginRelease, PluginReleaseProblem } from '../../types/plugin/plugin-release.js';
+import { format as f } from '../format.js';
+import { noteWrapped } from '../prompt.js';
+
+const PLUGIN_CONFIG_FILE = 'plugin-config.json';
+const NOTE_TITLE = 'printed, not executed';
```
mrafnadeem-apimatic:
```suggestion
const NOTE_TITLE = 'Commands to Copy and Run';
```

README.md
```diff
@@ -253,6 +254,33 @@ EXAMPLES
...
+## `apimatic plugin publish`
+
+Print the git commands for publishing your context plugin to GitHub.
+
+```
+USAGE
+  $ apimatic plugin publish [-i <value>] [-d <value>]
+
+FLAGS
+  -d, --destination=<value>  [default: <input>/plugin] path where the plugin will be generated.
```
mrafnadeem-apimatic: Should be "was", not "will be".
```suggestion
  -d, --destination=<value>  [default: <input>/plugin] path where the plugin was generated.
```

This should be fixed in the command code.

mrafnadeem-apimatic (changes requested):

mrafnadeem-apimatic (changes requested):
Do this in a new PR!: You should update `fileService.getDirectory` and `fileService.cleanDirectoryExcluding` to use a try-catch block and return a Result type. In case of failure, you would show a human-readable error message, and return a failed ActionResult.

src/types/plugin-context.ts
MuHamza30: removed the zip flag from plugin generate command

src/prompts/plugin/publish.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * The tag is annotated because `git push --follow-tags`, which ends both flows, carries annotated
+   * tags only and would leave a lightweight one behind on the machine.
+   */
+  private stageCommitAndTag(release: PluginRelease): string {
+    const label = `${release.toRepositoryName()} ${release.toVersion()}`;
+    return (
+      `${f.cmdAlt('git', 'add', '.')}\n` +
+      `${f.cmdAlt('git', 'commit', '-m', `"Publish ${label}"`)}\n` +
```
mrafnadeem-apimatic, line 101: You can remove `git add` here by just changing one character.

```suggestion
      `${f.cmdAlt('git', 'commit', '-am', `"Publish ${label}"`)}\n` +
```

src/types/flags-provider.ts
```diff
@@ -11,11 +11,11 @@ export class FlagsProvider {
...
-  public static destination(artifact: string, artifactName: string) {
+  public static destination(artifact: string, artifactName: string, tense: "will be" | "was" = "will be") {
     return {
       destination: Flags.string({
         char: "d",
-        description: `[default: <${FlagsProvider.inputFlagName}>/${artifact}] path where the ${artifactName} will be generated.`
+        description: `[default: <${FlagsProvider.inputFlagName}>/${artifact}] path where the ${artifactName} ${tense} generated.`
       })
     };
   }
```
mrafnadeem-apimatic: Instead of introducing a parameter here for the tense, it would be better to just inline this into the command itself and update the description there. The parameter is a risky change because wording can change in almost any way you can imagine. It's ok to have some duplication.

mrafnadeem-apimatic (changes requested):
see comments

mrafnadeem-apimatic (approved):

PR #329, opened 2026-08-27
fix: add docs link to publishing profile messages

mrafnadeem-apimatic:
@
Six publishing-profile messages told the user to go create or fix a profile "on the APIMatic App" without ever giving them an address. Only `publishing/profile/list.ts` linked anything at all.

## Changes

- New `src/prompts/publishing/links.ts` exporting `SDK_PUBLISHING_OVERVIEW_URL`, so the docs link has one home instead of being retyped per prompt.
- Six messages now append `Learn more: <link>`, with the existing sentences left verbatim:
  - `publishing/profile/list.ts` `noProfilesFound()` — its two older URLs and both module-level consts are replaced by the single overview link
  - `sdk/publish/interactive.ts` `noPublishingProfilesFound()` and `noProfileWithEnabledLanguagesFound()`
  - `sdk/publish/non-interactive.ts` `publishingProfileNotFound()`, `languageNotConfiguredForProfile()`, `publishTypesNotAvailableForLanguage()`
- `PublishingProfiles.create()` no longer returns user-facing copy from the types layer; `noPublishingProfilesFound()` owns its own text and no longer takes a string.

## Notes

- `commands/sdk/publish.ts` and `commands/publishing/profile/list.ts` are deliberately untouched — their descriptions mention profiles but never send the user anywhere.
- `pnpm test` currently fails at its `posttest` step (`eslint . --ext .ts`) because it crawls stale checkouts under `.claude/worktrees/`. That is pre-existing and unrelated; mocha passes.

## Verification

- `tsc -b` clean
- `eslint "src/**/*.{js,ts}"` clean
- mocha: 290 passing, 11 pending, 0 failing
- Messages rendered from the compiled output to confirm the link and clack gutter render correctly

@

### DIFF
--- a/src/actions/sdk/publish/interactive.ts
+++ b/src/actions/sdk/publish/interactive.ts
@@ -55,7 +55,7 @@ export class SdkPublishInteractiveAction {
     const publishingProfileItems = publishingProfilesResponse.value;
     const publishingProfilesResult = PublishingProfiles.create(publishingProfileItems);
     if (publishingProfilesResult.isErr()) {
-      this.prompts.noPublishingProfilesFound(publishingProfilesResult.error);
+      this.prompts.noPublishingProfilesFound();
       return ActionResult.failed();
     }
 
--- a/src/types/publish/publishing-profiles.ts
+++ b/src/types/publish/publishing-profiles.ts
@@ -9,9 +9,9 @@ import { PublishingProfile } from './publishing-profile.js';
 export class PublishingProfiles {
   private constructor(private readonly items: PublishingProfile[]) {}
 
-  public static create(items: PublishingProfileItem[]): Result<PublishingProfiles, string> {
+  public static create(items: PublishingProfileItem[]): Result<PublishingProfiles, 'No publishing profiles found.'> {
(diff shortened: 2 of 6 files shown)


src/types/publish/publishing-profiles.ts
mrafnadeem-apimatic, line 1: Now this error value isn't read by any code, but it does make the `create` function signature declarative.

src/actions/sdk/publish/interactive.ts
```diff
@@ -55,7 +55,7 @@ export class SdkPublishInteractiveAction {
     const publishingProfileItems = publishingProfilesResponse.value;
     const publishingProfilesResult = PublishingProfiles.create(publishingProfileItems);
     if (publishingProfilesResult.isErr()) {
-      this.prompts.noPublishingProfilesFound(publishingProfilesResult.error);
+      this.prompts.noPublishingProfilesFound();
```
mrafnadeem-apimatic, line 58: now the noPublishingProfilesFound method is responsible for the actual message

saeedjamshaid (approved):

PR #335, opened 2026-09-01
fix: drop Claude Code references from plugin generate

sohail2721:
`plugin generate` now emits Cursor and VS Code manifests alongside the Claude Code one, so singling out Claude Code in the command's user-facing text is misleading.

## Changes

- `src/commands/plugin/generate.ts` — summary is now "Generate a context plugin for your published SDKs." (the longer description was already editor-agnostic).
- `src/prompts/plugin/generate.ts` — removed `tryPluginInClaudeCode`, which printed `claude --plugin-dir "<path>"`. There is no single command that covers all three editors, so the message is dropped rather than genericized; `pluginGenerated` still prints where the artifacts landed and `nextStepsPublishPlugin` still follows.
- `src/actions/plugin/generate.ts` — dropped the call site.
- `README.md` — regenerated command summary lines to match.

No other prompt in the command mentions an editor. `npm run build`, `npm run lint`, and `npm test` (291 passing) are green; no test asserted on the removed prompt.

https://claude.ai/code/session_01Xfvc6A87iGFUj2FwtTvQQK

### DIFF
--- a/src/actions/plugin/generate.ts
+++ b/src/actions/plugin/generate.ts
@@ -90,7 +90,6 @@ export class PluginGenerateAction {
       await pluginContext.save(tempPluginZipPath);
 
       this.prompts.pluginGenerated(pluginDirectory);
-      this.prompts.tryPluginInClaudeCode(pluginDirectory);
       this.prompts.nextStepsPublishPlugin();
 
       return ActionResult.success();
--- a/src/actions/plugin/generate.ts
+++ b/src/actions/plugin/generate.ts
@@ -90,6 +90,7 @@ export class PluginGenerateAction {
       await pluginContext.save(tempPluginZipPath);
 
       this.prompts.pluginGenerated(pluginDirectory);
+      this.prompts.tryPluginLocally(pluginDirectory);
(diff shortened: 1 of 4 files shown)


src/actions/plugin/generate.ts
```diff
@@ -90,7 +90,6 @@ export class PluginGenerateAction {
       await pluginContext.save(tempPluginZipPath);
 
       this.prompts.pluginGenerated(pluginDirectory);
-      this.prompts.tryPluginInClaudeCode(pluginDirectory);
```
mrafnadeem-apimatic, line 93: Instead of deleting, we should guide the user on how to install it for different IDEs. Check the context-plugin-installer CLI as well if it supports local plugin folder installation.

mrafnadeem-apimatic (changes requested):

src/actions/plugin/generate.ts
mrafnadeem-apimatic: To make things easier, you can link to official documentation of each IDE.

mrafnadeem-apimatic (approved):

PR #337, opened 2026-09-02
feat: offer typescript v4 codegen in interactive sdk publish

mehnoorsiddiqui:
## What

Registers `V4 (beta)` as a code generator option for TypeScript in `CODEGEN_OPTIONS`.

```diff
-  [Language.TYPESCRIPT]: [CodegenOption.v3]
+  [Language.TYPESCRIPT]: [CodegenOption.v3, CodegenOption.create(CodeGenerationVersion.V4, Stability.BETA)]
```

## Why

Enabling TypeScript V4 for context plugins comes down to letting a TypeScript SDK be published as V4 in the first place. Tracing the plugin path: `plugin generate` zips `src/` and uploads it with `plugin-config.json`, and neither `PluginGenerateAction` nor `PluginService` filters by language or codegen version. The `codegenVersion` on each language entry is whatever `sdk publish` recorded through `PluginRecordSdkAction`. `CODEGEN_OPTIONS` is the only CLI-side gate.

The interactive flow shows the picker only when a language has more than one option:

```ts
const codegenOptions = getCodegenOptions(language);
const codegenOption = codegenOptions.length === 1 ? codegenOptions[0]
  : await this.prompts.selectCodegenVersion(codegenOptions);
```

TypeScript had one, so an interactive publish was silently pinned to V3 and always wrote `codegenVersion: "v3"`. It now offers `V3 (stable)` / `V4 (beta)`, and the selection flows into both `generateV4Sdk` and the plugin-config entry.

Non-interactive `--language typescript --codegen-version v4 --stability beta` already parsed — validation there is server-side, as asserted by the existing test "does not filter languages by codegen version, leaving that to the service". Only the interactive path was gated.

Beta matches how C# V4 is registered today.

## Tests

Adds `test/types/sdk/generate.test.ts`:

- TypeScript offers V3 and V4
- TypeScript V4 is beta
- V3 stays first for every language, since the picker's `initialValue` is `options[0]`

Full suite: 294 passing, lint clean.

### DIFF
--- a/src/types/sdk/generate.ts
+++ b/src/types/sdk/generate.ts
@@ -99,7 +99,7 @@ export const CODEGEN_OPTIONS: Readonly<Record<Language, Readonly<NonEmptyArray<C
   [Language.PHP]: [CodegenOption.v3],
   [Language.PYTHON]: [CodegenOption.v3],
   [Language.RUBY]: [CodegenOption.v3],
-  [Language.TYPESCRIPT]: [CodegenOption.v3]
+  [Language.TYPESCRIPT]: [CodegenOption.v3, CodegenOption.create(CodeGenerationVersion.V4, Stability.BETA)]
 };
 
 export function getCodegenOptions(language: Language): Readonly<NonEmptyArray<CodegenOption>> {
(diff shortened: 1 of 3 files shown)


mrafnadeem-apimatic:
Closed until package publishing is verified.

PR #340, opened 2026-09-09
fix: stop command suggestions offering an unrunnable command and looping

Darker98:
Fixes apimatic/apimatic-io#2214

## Problem

`apimatic auth help` warned, suggested `help auth`, and accepting that suggestion re-offered the *same* suggestion forever. The only exit was `n`, which then pointed at the malformed `apimatic help help`.

```
» Warning: auth help is not a apimatic command.
Did you mean help auth? (Y/n) Y
» Warning: help auth is not a apimatic command.
Did you mean help auth? (Y/n) Y          <- identical, forever
» Error: Run apimatic help help for a list of available commands.
```

## Root cause

All three symptoms come from `src/hooks/not-found.ts`, a fork of `@oclif/plugin-not-found`.

1. **The suggestion was not a command.** The help special-case built the *string* `"help:auth"`. Upstream then rewrites it (`suggestion = "help"`, `argv = ["auth"]`) and runs the rewritten value; the fork captured `confirmedSuggestion` **before** that rewrite and ran `"help:auth"` — not a command id.
2. **The loop.** `config.runCommand("help:auth")` therefore fell through to `command_not_found` again with id `help:auth`, whose help special-case regenerated the identical `help:auth`. No state, no guard, so `Y` could never converge.
3. **`apimatic help help`.** `config.findTopic("help")` is truthy — oclif registers a topic entry for *every* command id, leaf commands included — so the first segment got appended to the hint.

## Changes

- Model a suggestion as `{ id, argv }` and validate it with `findCommand` before prompting, so an unrunnable suggestion is never offered. This removes the bug class by construction.
- Guard re-entry with a single retry, then report the failure instead of re-prompting.
- Narrow the fallback hint only on a topic that is **not itself a command**, so `help zzz` gives `apimatic help` while `sdk genrate` still gives the useful `apimatic help sdk`.
- Fix the article: *"is not **an** apimatic command"*.

## Behaviour

| Input | Before | After |
|---|---|---|
| `auth help` | loops, then `apimatic help help` | warns, offers a runnable match, `apimatic help auth` on decline |
| `help zzz` | `apimatic help help` | `apimatic help` |
| `sdk genrate` | prompt | prompt → `apimatic help sdk` on decline |

## Not included

The issue's last bullet — *ideally `apimatic auth help` is simply treated as `apimatic help auth`* — is deliberately left out of this PR. It is a UX nicety rather than part of the defect, and is easy to add separately. The loop, the unrunnable suggestion, and the bad hint are all fixed without it.

## Testing

New `test/hooks/not-found.test.ts` (6 tests) pins each behaviour against a real loaded `Config`. Full suite: **297 passing**; typecheck and lint clean.

Worth noting: running these tests against the *old* hook **hangs until killed** — the recursion test reproduces the infinite loop in-process.

### DIFF
--- a/src/hooks/not-found.ts
+++ b/src/hooks/not-found.ts
@@ -4,6 +4,10 @@ import { cyan, yellow } from "ansis";
 
 import utils from "./utils.js";
 
+// The hook re-enters through `config.runCommand`, so a suggestion that fails to resolve
+// would prompt forever. Allow a single retry per process, then report the failure.
+let retrying = false;
+
 const hook: Hook.CommandNotFound = async function (opts) {
   const hiddenCommandIds = new Set(opts.config.commands.filter((c) => c.hidden).map((c) => c.id));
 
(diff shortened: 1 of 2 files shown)


src/hooks/not-found.ts
```diff
@@ -15,23 +19,22 @@ const hook: Hook.CommandNotFound = async function (opts) {
...
     binHelp = `${binHelp} ${idSplit[0]}`;
   }
 
-  let suggestion: string | null;
-  if (/:?help:?/.test(opts.id)) {
-    suggestion = ["help", ...opts.id.split(":").filter((cmd) => cmd !== "help")].join(":");
-  } else {
-    suggestion = utils.closest(opts.id, commandIDs);
-  }
+  const suggestion = utils.closest(opts.id, commandIDs);
```
sohail2721, line 28: Dropping the help branch makes topics unsuggestable, so apimatic auth help now offers auth login behind a default-yes prompt and starts an interactive login; a suggestion needs to carry (id, argv) so a help request can resolve as help + ["auth"],  which also keeps the loop fixed, since findCommand("help") resolves.

test/hooks/not-found.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    expect(confirm.callCount, "should prompt at most once").to.equal(1);
+  });
+
+  it("points at `apimatic help`, never `apimatic help help`", async () => {
+    sandbox.stub(utils, "getConfirmation").resolves(false);
+
+    await expectListHint("help:zzz");
+
+    expect(errors[0]).to.contain("apimatic help ");
```
sohail2721, line 99: contain("apimatic help ") has a trailing space that cyan.bold's ANSI reset sits inside, so this fails on any color-capable terminal and only passes under NO_COLOR=1; drop the space or strip ANSI before asserting.

sohail2721 (changes requested):

PR #343, opened 2026-09-17
feat(portal): build the documentation portal locally with Fumadocs

saeedjamshaid:
## Summary

Replaces the hosted portal generation with a portal built on the user's machine. `portal generate` now copies a Fumadocs / TanStack Start template into a temp project, points it at the user's `src/`, runs Vite, and writes a fully static site. `portal serve` runs the same template on Vite's dev server with live reload. Nothing is uploaded to build a portal any more; the only network call is the entitlement check.

Design and decisions: `.ai/plans/fumadocs-portal.md`. Section 12 lists what remains after this PR.

## What changes for users

- **New input layout**, additive: `src/portal.json` (title, description, logo, siteUrl), `src/spec/` (shared with SDK generation), `src/content/` (Markdown with front matter, ordered by `meta.json`), `src/static/`. `APIMATIC-BUILD.json` stays for the SDK and plugin commands only. Both commands print a migration hint when they find `generatePortal` in the build file and no `portal.json`.
- **Removed**: `portal toc new`, `portal recipe new`, `portal copilot`. Hidden stubs remain for one major, explain what replaced them, and exit 1.
- **`portal serve`** drops `--destination` and `--no-reload`; reload is always on.
- **Node 22.12 or newer** is required to build a portal. The engine field says so and the commands check before building.
- **Quickstart** scaffolds the new layout and downloads its default spec from the sample repository's permanent `v2` branch.
- Request samples: curl only. Other languages will come from `x-codeSamples` in the spec.

## How it is built

- `portal-template/` ships in the package. Per-dependency junctions or symlinks link the CLI's own installed packages into the temp project, so users install nothing extra.
- SPA mode with an explicit prerender page list: every URL is complete HTML before JavaScript runs, then hydrates. Per-page title, description and canonical link; sitemap and robots when `siteUrl` is set; Orama search index; `llms.txt` and a `.md` twin of every page.
- Modules that read the spec off disk live in `*.server.ts` files; TanStack import protection fails the build if client code imports one. This closes a bug where the loader shipped to the browser and left every page inert.
- Each operation page carries only its own path item and the components it references; the sidebar tree is loaded once from the root route.

## Verification

- 328 tests locally on Windows / Node 23.4, including the gated end-to-end build (`APIMATIC_E2E=1`) and the new suites for the config, source context, authorization gate, project service, template packaging, and the document trim.
- Generated pages and `portal serve` verified in headless Chrome: no console errors, code highlighting and interactive controls appear after hydration.
- Whole-spec builds measured (Petstore 12.6 MB; Stripe 594 ops 2.2 GB in 375 s; GitHub 1,511 pages 2.4 GB in 229 s). For large specs the sidebar tree, repeated in every page, dominates output; noted as the next follow-up in the plan.
- **Not yet run**: the new `Tests` workflow's ubuntu / windows / macos x Node 22.12 / 24 matrix. This PR is what triggers it for the first time.

## Squash commit footer

The squash commit needs this footer so semantic-release cuts the major:

```
BREAKING CHANGE: building a portal requires Node >= 22.12 and the new src/portal.json layout.
portal toc new, portal recipe new and portal copilot are removed (hidden stubs exit 1);
portal serve drops --destination and --no-reload. Run `apimatic autocomplete --refresh-cache`
to drop the removed commands from shell completion.
```

## Known and accepted for this cut

- `$ref` values in a spec are read by Fumadocs' bundler during the build, including file paths and URLs (plan, section 9).
- Generated sites load Geist from Google Fonts and the page actions link to ChatGPT, Claude and Cursor.
- The runtime set is pinned exactly; no Renovate or Dependabot config exists yet to move it.

https://claude.ai/code/session_01MfFVFknGtZmWnwUf1ntQrE


src/infrastructure/services/sdk-generation-service.ts
mehnoorsiddiqui, line 1: portal-service.ts is down to generateSdk and generateV4Sdk, imported only by sdk/generate.ts. The name is now actively misleading, and this PR is the natural moment to rename it (SdkGenerationService) after the next few PRs it becomes a much bigger diff.

src/infrastructure/portal-project-service.ts
```diff
@@ -0,0 +1,213 @@
...
+      viteBinary: new FilePath(viteDirectory.join('bin'), new FileName('vite.js'))
+    });
+  }
+
+  /**
+   * Environment for the child process. Deliberately an allow-list: the auth key must not
+   * reach the build, and any stray `VITE_*` variable would be inlined into the output.
+   */
+  public childEnvironment(): Record<string, string> {
+    const allowed = ['PATH', 'Path', 'HOME', 'USERPROFILE', 'TEMP', 'TMP', 'SYSTEMROOT', 'SystemRoot', 'SystemDrive'];
```
mehnoorsiddiqui: it allows TEMP/TMP (Windows spellings) but not the POSIX TMPDIR, so on macOS/Linux the child's os.tmpdir() silently falls back to /tmp. LANG/LC_ALL are also absent. The allow-list approach is right , the list is just incomplete.

src/infrastructure/portal-build-service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const pageCount = await this.countPages(output);
+    if (pageCount === 0) {
+      return err({ message: 'The portal build produced no pages.', log });
+    }
+
+    return ok({ output, pageCount });
+  }
+
+  private async countPages(output: DirectoryPath): Promise<number> {
```
mehnoorsiddiqui: "Actions return ActionResult… No uncaught throws above infrastructure." This in portal generate break that :

countPages → getDirectory stats every entry recursively and throws on any it can't read. It runs after a successful build, so the rejection propagates through withSpinner (/src/prompts/prompt.ts), which await fns without a catch. Result: the clack spinner is left spinning and an oclif stack trace prints.

src/types/portal-context.ts
```diff
@@ -1,44 +1,55 @@
...
   }
 
   public async exists() {
-    return !await this.fileService.directoryEmpty(this.portalDirectory);
+    return !(await this.fileService.directoryEmpty(this.portalDirectory));
   }
 
-  public async save(tempPortalFilePath: FilePath, zipPortal: boolean) {
+  /** Writes the finished site to the portal directory, as files or as a single archive. */
+  public async save(builtDirectory: DirectoryPath, asZip: boolean) {
```
mehnoorsiddiqui: "Actions return ActionResult… No uncaught throws above infrastructure." This in portal generate break that :
`save() `calls cleanDirectory(portalDirectory) and then copyDirectoryContents. If the copy fails partway (ENOSPC on a 2 GB portal is not hypothetical given your own Stripe/GitHub measurements, or EACCES), the user's previous portal is already gone and the new one is incomplete. It also throws raw, outside any spinner.

`saveBuildLog `on the failure path has the same exposure.

test/e2e/portal-build.test.ts
```diff
@@ -0,0 +1,144 @@
...
+(enabled ? describe : describe.skip)('portal build (end to end)', function () {
+  this.timeout(10 * 60 * 1000);
+
+  const fixture = new DirectoryPath(process.cwd()).join('test/resources/portal-inputs/default');
+
+  let root: string;
+  let project: DirectoryPath;
+  let output: DirectoryPath;
+
+  before(async () => {
```
mehnoorsiddiqui, line 27: calls ensureBuildDirectoryBase(fixture) directly rather than withBuildDirectory, and after() removes only root. On the Windows runners (workspace D:, temp C:) this leaves test/resources/portal-inputs/.apimatic-build behind — the exact git status pollution [tmp-extensions.ts:69-74]was written to prevent.

src/types/portal-source-context.ts
```diff
@@ -0,0 +1,275 @@
...
+import { FilePath } from './file/filePath.js';
+import { OpenApiDocument } from './portal/openapi-document.js';
+import { PortalConfig } from './portal/portal-config.js';
+import { PortalMigration, PortalSource, PortalSourceProblem, PortalSpec } from './portal/portal-source.js';
+import { SpecContext } from './spec-context.js';
+import { stripByteOrderMark } from '../utils/string-utils.js';
+
+const SPEC_EXTENSIONS = ['.json', '.yaml', '.yml'];
+
+// Empty on purpose: the portal's own routes under /api/ are files with extensions --
```
mehnoorsiddiqui, line 16: Content pages under /api/ can collide with spec sections. This comment reasons that portal routes under /api/ are files with extensions, so nothing can collide — true for /api/search.json, but a user's content/api/foo.md and a spec slug foo both resolve to /api/foo, and the merged loader picks one silently. Cheap to detect in resolve() and warn.

src/types/portal-source-context.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  private async readDocument(file: FilePath): Promise<OpenApiDocument | undefined> {
+    try {
+      return OpenApiDocument.parse(file.name(), await this.fileService.getContents(file));
+    } catch {
+      return undefined;
+    }
+  }
+
+  // A document without a version key is not a spec at all (APIMATIC-META.json, a `$ref`
```
mehnoorsiddiqui: the comment above uniqueSlug describes what specs() does, not what uniqueSlug does.

src/types/portal-source-context.ts
```diff
@@ -0,0 +1,275 @@
...
+import { DirectoryPath } from './file/directoryPath.js';
+import { FileName } from './file/fileName.js';
+import { FilePath } from './file/filePath.js';
+import { OpenApiDocument } from './portal/openapi-document.js';
+import { PortalConfig } from './portal/portal-config.js';
+import { PortalMigration, PortalSource, PortalSourceProblem, PortalSpec } from './portal/portal-source.js';
+import { SpecContext } from './spec-context.js';
+import { stripByteOrderMark } from '../utils/string-utils.js';
+
+const SPEC_EXTENSIONS = ['.json', '.yaml', '.yml'];
```
mehnoorsiddiqui, line 14: Are we not supporting other extensions like wsdl  etc ..?

portal-template/src/components/api-page.tsx
mehnoorsiddiqui, line 1: portal-template is fully eslint-ignored

src/actions/portal/quickstart.ts
```diff
@@ -56,29 +53,22 @@ export class PortalQuickstartAction {
...
-      if (!accountInfo.value.isOnPremGenerationAllowed) {
-        this.prompts.onPremGenerationNotAllowedOnPlan();
-        return ActionResult.cancelled();
-      }
-      const allowedLanguages = mapLanguages(accountInfo.value.allowedLanguages);
+    // Checked before any question is asked: `portal serve` refuses without this entitlement.
+    const authorization = await this.authorizationService.authorize(this.configDir, this.commandMetadata.shell, null);
+    if (authorization.isErr()) {
+      this.prompts.authorizationFailed(authorization.error);
+      return ActionResult.cancelled();
```
mehnoorsiddiqui: [quickstart.ts:60] returns ActionResult.cancelled() for a denied entitlement — exit 130, which [action-result.ts:2] reserves for Ctrl+C. src/actions/portal/generate.ts#L57 and src/actions/portal/serve.ts#L52 return failed() (exit 1) for the identical PortalAuthorizationFailure. Any script branching on exit code reads an entitlement denial as a user cancellation.

src/types/portal/portal-config.ts
```diff
@@ -0,0 +1,250 @@
...
+      ...PortalConfig.descriptionErrors(description),
+      ...PortalConfig.logoErrors(logo),
+      ...PortalConfig.aiPageActionsErrors(aiPageActions),
+      ...PortalConfig.siteUrlErrors(siteUrl, parsedSiteUrl)
+    ];
+    if (errors.length > 0) {
+      return err(errors);
+    }
+
+    // A blank description is the same as none; without this it shipped as the site
```
mehnoorsiddiqui: titleErrors and friends return error arrays without narrowing types, so the constructor call re-asserts what validation just proved — (title as string).trim(), (aiPageActions as boolean | undefined) ?? true. Weaken a validator and that cast throws at runtime with no compile error. Having each validator return a Result<T> would make the guarantee structural.

src/actions/portal/quickstart.ts
```diff
@@ -1,53 +1,50 @@
...
-import { ApiService } from '../../infrastructure/services/api-service.js';
-import { DEFAULT_COPILOT_WELCOME_MESSAGE } from './copilot.js';
-import { mapLanguages } from '../../types/sdk/generate.js';
+import { OpenApiDocument, SpecFormat } from '../../types/portal/openapi-document.js';
+import { PortalSourceContext } from '../../types/portal-source-context.js';
+import { PortalAuthorizationService } from '../../infrastructure/services/portal-authorization-service.js';
+import { FileDownloadService } from '../../infrastructure/services/file-download-service.js';
+import { PortalProjectService } from '../../infrastructure/portal-project-service.js';
 
 const defaultPort: number = 23513 as const;
```
mehnoorsiddiqui: Port 23513 hardcoded in four places across src/actions/portal/quickstart.ts#L20 and src/commands/portal/serve.ts#L28. The declaration const defaultPort: number = 23513 as const also has a no-op as const — the : number annotation widens it straight back.

src/actions/portal/serve.ts
mehnoorsiddiqui, line 1: generate.ts:45-72 and serve.ts:36-67 share ~35 lines in the same order: runtime gate → authorize → resolve source → report shadowed files → withBuildDirectory → prepare. The quickstart action repeats the runtime gate too. Also duplicated verbatim in both commands: const sourceDirectory = input ? new DirectoryPath(input, 'src') : workingDirectory.join('src') — where both branches compute the same path, since DirectoryPath resolves in its constructor.

src/types/portal/portal-config.ts
```diff
@@ -0,0 +1,250 @@
...
+    return this.description;
+  }
+
+  /** Whether each page offers to open itself in an external AI assistant. */
+  public offersAiPageActions(): boolean {
+    return this.aiPageActions;
+  }
+
+  /** Path of the logo relative to `src/` with forward slashes, or null when none is configured. */
+  public logoPath(): string | null {
```
mehnoorsiddiqui, line 192: returns raw string . Related: logoSiteUrl() does substring(STATIC_PREFIX.length - 1) — index arithmetic that'll break quietly if the prefix changes.

portal-template/src/lib/openapi.server.ts
```diff
@@ -0,0 +1,21 @@
...
+export type OpenApiSource = StaticSource<{ pageData: OpenAPIPageData; metaData: MetaData }>;
+
+/**
+ * Every document's reference pages as one source, each section mounted under `api/<slug>`.
+ * One server per document, because `staticSource()` emits pages for every schema its server
+ * knows about, so sharing a server across sections duplicates pages. Merged afterwards so a
+ * reference page has the same type whichever document it came from, which is what lets the
+ * routes tell it apart from a Markdown page.
+ */
+export async function openApiSource(): Promise<OpenApiSource> {
```
mehnoorsiddiqui, line 13: Prerender and runtime construct the OpenAPI source twice. portal-template/src/lib/openapi.server.ts#L17) and portal-template/prerender-pages.ts#L70-L71 both call createOpenAPI({...}).staticSource({ baseDir: \api/${id}`, groupBy: 'tag', meta: true }). Change baseDirorgroupBy` in one and the prerender list silently desynchronises from the real routes — pages 404 with no build error. This is the duplication I'd extract first.

src/infrastructure/services/sdk-generation-service.ts
saeedjamshaid: Done. Renamed to `SdkGenerationService` and the file to `sdk-generation-service.ts`, with the test file renamed to match.

src/infrastructure/portal-project-service.ts
saeedjamshaid: Good catch. The allow-list now passes `TMPDIR`, `LANG` and `LC_ALL` as well.

src/infrastructure/portal-build-service.ts
saeedjamshaid: Fixed. `countPages` returns null when the output cannot be read and the build reports it as a failure through the normal Result path, so the spinner stops cleanly.

src/types/portal-context.ts
saeedjamshaid: Reworked. The site is staged inside the destination first, so a failing copy leaves the previous portal exactly as it was. If the swap itself fails, whatever could not be moved into place stays in the staging folder and the message says where it is; a leftover staging folder also counts as an existing portal, so the next run asks before touching it. `saveBuildLog` no longer throws either. Both paths have tests.

test/e2e/portal-build.test.ts
saeedjamshaid: Fixed. The test now removes the fallback base after the run, and the cleanup helper refuses to touch the system temp directory.

src/types/portal-source-context.ts
saeedjamshaid: Added. `resolve()` now reports spec slugs that a content page also claims at `/api/<slug>`, and both commands warn. The addresses come from fumadocs' own `getSlugs`, so `(group)` folders and `index` collapsing follow the same rules as the site, and the comparison ignores case since the prerender writes both to one path on a case-insensitive disk. The comment about the portal's own routes was clarified too.

src/types/portal-source-context.ts
saeedjamshaid: Moved to the skip logic in `specs()` that it describes.

src/types/portal-source-context.ts
saeedjamshaid: Not for the local build. The portal renders with fumadocs-openapi, which reads OpenAPI 3.x only. Other formats are reported as unsupported with a pointer to `apimatic api transform`, which converts them first.

portal-template/src/components/api-page.tsx
saeedjamshaid: Intentional for now. The template is React/TSX built by Vite, so the CLI's Node config does not apply, and it is type-checked in the e2e build against the packages it ships with. A separate React lint config needs new dev dependencies, so I would take that as a follow-up.

src/actions/portal/quickstart.ts
saeedjamshaid: Fixed, quickstart now returns `failed()` like generate and serve.

src/types/portal/portal-config.ts
saeedjamshaid: Refactored. Each validator returns a typed `Result`, combined with `Result.combineWithAllErrors`, so the constructor is fed the proven values and the casts are gone.

src/actions/portal/quickstart.ts
saeedjamshaid: Consolidated into `DEFAULT_PORTAL_PORT` exported by the serve action, and the no-op `as const` is gone.

src/actions/portal/serve.ts
saeedjamshaid: The duplicated `sourceDirectory` line is simplified in both commands. Extracting the shared prologue is deferred to a follow-up so this PR stays reviewable.

src/types/portal/portal-config.ts
saeedjamshaid: Replaced with a slice past the prefix. `logoPath()` stays a string because it feeds the `PortalIdentity` JSON handed to the browser bundle, which is a string boundary.

portal-template/src/lib/openapi.server.ts
saeedjamshaid: Extracted to `openApiSection()` in `src/lib/openapi-section.server.ts`, used by both the site source and the prerender list.

mehnoorsiddiqui (approved):

PR #346, opened 2026-09-22
feat(portal): order the sidebar from nav.json

saeedjamshaid:
Gives users control over portal sidebar order and folder names through a `nav.json` beside
their pages, and gives the API reference a sensible default shape. Implements
`.ai/plans/portal-navigation.md`, which follows on from #343.

## The format

One `nav.json` per content directory, with two fields:

```json
{ "title": "Developer Guides", "pages": ["index", "authentication", "apimatic:api", "..."] }
```

`title` names the folder the file sits in, and outranks both the directory name and the title
Fumadocs takes from an index page. It is refused in two places where it would set nothing: the
content root's own file, which orders no folder — the portal's own name is `title` in
`portal.json` — and a directory with no page beneath it, which becomes no folder at all. It
names a folder rather than moving it: a folder with a `title` and no entry positioning it still
sorts where its directory name put it, because Fumadocs has already ordered the children by the
time a transformer sees them. `pages` is how you move things.

Entries in `pages` name a page (without its extension) or a subfolder in the same directory.
Three tokens are recognised, and the two `apimatic:` ones only at the top of the content
directory, because that is where the nodes they position live:

| Token | |
|---|---|
| `...` | Everything in this directory no other entry names |
| `apimatic:api` | The whole API reference, as one node |
| `apimatic:pages` | The pages the CLI generates and injects (none yet; the SDK page follows) |

Anything not named keeps its default order rather than disappearing, and a file that names
something that does not exist **fails the build naming the entry** — Fumadocs drops such an
entry silently, which is the main reason the CLI validates the file itself.

## How it works

Ordering is a page-tree transformer (`portal-template/src/lib/navigation.ts`) rather than
resolution in the CLI. The hook runs per directory once Fumadocs has already built and ordered
its children, so it only ever permutes a list — losing a page is structurally impossible — and
editing `nav.json` reorders a running `portal serve` with no restart.

The metadata collection is restricted to `files: ['**/nav.json']`. Without it a leftover
`meta.json` is loaded as Fumadocs folder metadata and applied *before* any transformer runs,
and since a metadata file hides whatever it does not name, the transformer could not put those
pages back. The CLI warns about any `meta.json` it finds and tells the user to rename it.

## API reference

`api` would render as "Api", which reads as a typo, so the wrapper is titled "API Reference" —
a default, so a `title` in `content/api/nav.json` overrides it like any other folder's.

**`api` and `apimatic:api` both position the reference, and they position the same node.** It is
mounted at `content/api` in every portal, whether or not such a directory exists, and a
`content/api/` the user keeps for their own pages merges into that folder rather than making a
second one — so naming it `api` is fair, and listing both spellings names one node twice and is
refused. The mount point is a child of the content root in every portal, so the entry reads the
same way with a directory there and with nothing there.

A page called `api.md` at the content root is the one refusal: it is a second child that an
entry can never reach, because an entry resolves to a folder before a page, so the sidebar would
order the reference while the file reads as though it ordered the page. The error says to rename
the page.

Below the reference, a single specification's section only restates the portal's own title, so that level
is lifted away and the tag folders sit directly under the reference. Adding a second document
inserts the level back rather than renaming anything. Page addresses come from slugs, not tree
position, so no operation URL moves.

## Worth a reviewer's attention

- **The transformer leans on three things Fumadocs does not document as public API**: `$ref`
  surviving on tree nodes, `builder.resolveFlattenPath` being reachable from a transformer
  context, and `storage.read(path).type` carrying the key a source was passed to `loader()`
  under. All three are covered by the end-to-end sidebar assertions, which is what would catch
  a Fumadocs upgrade taking one away; `fumadocs-core` is pinned exactly, so such an upgrade is
  always a deliberate change here rather than drift.
- **Fumadocs' other folder-metadata keys are not settable**, and the narrowing is narrower
  than it first looked. `nav.json` is never read as folder metadata at all — a folder's
  metadata is looked up as `meta.meta`, and this file indexes as `nav.meta` — so restricting
  the collection does not "drop" individual keys; it means no folder has Fumadocs metadata.
  Everything `nav.json` does, it does itself. `title` is therefore implemented in the
  transformer rather than passed through, which is why it works; `icon`, `defaultOpen`,
  `collapsible`, `root` and `pagesIndex` have no equivalent and are named in the error rather
  than answered with a spelling. Nothing released depended on them.
- **A built portal publishes the content directory's absolute path** in `assets/dist-*.js`.
  `defineDocs({ dir })` compiles it in as `base`, and `src/lib/source.ts` cannot move behind
  `.server` because the browser imports it to lazy load page bodies. The specification paths
  and the project directory do *not* leak, and that is asserted; this one is a pending test
  and question 3 of the plan's section 11. Pre-existing, but this branch is what makes it
  visible — say if it should block the merge.
- **`pnpm readme` is broken on every platform, and the cause is repo-wide.** `README.md` is
  committed as CRLF — `git ls-files --eol` reports `i/crlf` — so it checks out that way on
  Linux and macOS too, and oclif's marker regex does not match across it: it appends duplicate
  usage and command sections instead of replacing them. Five other files are in the same state,
  including `.github/workflows/release.yml`. There is no `.gitattributes` at all; `* text=auto`
  with `git add --renormalize .` fixes all of it. The README here was generated by converting
  to LF and back. Not done in this PR: it is repo infra, and it rewrites files this branch does
  not otherwise touch.

## Testing

601 unit tests and 18 end-to-end (1 pending, above). The end-to-end suite is the only place
the Vite glob, the collection restriction and the transformer are exercised over real on-disk
storage, and the only place the template is type-checked against the packages it ships with.
The fixture's `content/guides/` is named by its `nav.json` rather than by its directory, so
the title is proven through a real build and not only through the in-memory loader.

Verified by hand against a running `portal serve`: reordering pages and moving `apimatic:api`
both take effect without a restart; adding or removing a page still does not, which is what
the command's description now says.


portal-template/src/lib/navigation.ts
mehnoorsiddiqui, line 1: github is showing red cross on this file 

Image

src/types/portal/portal-navigation.ts
mehnoorsiddiqui, line 1: red cross on the file by github

src/types/portal-source-context.ts
mehnoorsiddiqui, line 1: red cross on the file by github

src/types/portal/portal-navigation.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    // A page and a folder of one name are both children, and an entry positions the folder,
+    // as it does in Fumadocs' own metadata. The page could then never be positioned, which
+    // is the quietly wrong sidebar this file exists to refuse.
+    if (context.childNames.filter((name) => name === entry).length > 1) {
+      return err(
+        `${context.label}: '${entry}' is both a page and a folder in this directory, and the entry ` +
+          `positions the folder. Rename the page to position it.`
+      );
+    }
```
mehnoorsiddiqui, line 151: `api` at the content root can mean two things, and the two halves of the format disagree about which.

With `content/api.md` and **no** `content/api/` directory, `childNames` contains `api` (the page), so this check passes. But the template resolves an entry to a folder first (`navigation.ts` L216-224), and there *is* a folder at `api` — the API reference. So the entry positions the reference, and the user's page falls into the unnamed band:

```json
{ "pages": ["index", "api", "authentication"] }
```

gives `Welcome, API Reference, Authentication, My API notes` — verified against a real `loader()` — with no error from the CLI. That is the quietly wrong sidebar this file exists to refuse.

The rule at L171 is already the right message for it; it just never fires, because the name appears once — the reference folder is never added to `childNames`. The sibling case (`api.md` *and* `api/`) is caught correctly.

Suggestion: thread the root's page names into `NavigationContext` and refuse `api` at the content root when it is a page — e.g. "`'api'` is the directory the API reference is mounted at, so a page of that name cannot be positioned. Rename the page." Worth a sibling case to `portal-navigation.test.ts` L113, which covers the same `childNames` assuming the name came from a folder.

.ai/plans/fumadocs-portal.md
```diff
@@ -65,19 +68,29 @@ for sectioning and are not copied into the output. Swagger 2.0 and
...
-from `title` and `description` so `/` always resolves. Folders without
-`meta.json` are ordered alphabetically by Fumadocs.
-
-Sidebar order: content pages in `meta.json` order first, then one section per
-spec file in filename order, grouped by tag inside each spec.
+The quickstart scaffolds a `content/index.md`, but nothing regenerates one: a
+portal whose content directory has no index page has no `/`. This paragraph
+promised a generated fallback home page and it was never built; the promise is
+dropped rather than restated, since `/` resolving is worth its own decision
+rather than a line in a layout section. Folders without `nav.json` are ordered
```
mehnoorsiddiqui: This looks wrong — the fallback home page was built in #343 and is already on `dev`. `src/routes/$.tsx` L70-74 returns a generated landing page when there is no index page ("A project without content/index.md(x) still gets a landing page"), and `prerender-pages.ts` L15 always seeds `/`, so the static build emits an `index.html` either way.

Worth correcting: a plan saying a feature does not exist is the kind of note someone acts on. If the point is that the fallback is *minimal*, that is fair — just a different sentence.

mrafnadeem-apimatic (approved):

PR #347, opened 2026-09-22
build!: move to Node 24 and TypeScript 7

mrafnadeem-apimatic:
Raises the engine floor to Node 24 and moves compilation to TypeScript 7, ahead of the next major.

## Why Node 24

Not because Node 24 is newest — because we should not ship against a runtime that reaches end of life inside this major's supported life. **Node 22 is EOL 2027-04-30**; Node 24 is patched until **2028-04-30**. That is the claim that survives a security questionnaire, and it is the reason worth quoting.

Two things to keep in view:

- Node 22 is still supported upstream for another seven months, so this **does** lock out anyone standardised on it. That is deliberate — a major is where a floor moves — but it is a real cost, not a free one.
- Node 24 enters *maintenance* on 2026-10-20 and Node 26 becomes LTS on 2026-10-28, both about five weeks out. The floor is the minimum we support, not what we recommend; 26 is what most developers will be running, which is why CI now tests both.

## Why TypeScript 7, and why TypeScript 6 is still here

Measured on this repo, cold, Windows, Node 24.12.0:

| Gate | TS 5.9.3 | TS 7.0.2 | |
|---|---|---|---|
| `build` (full emit) | 4939 / 4749 / 4534 ms | 1288 / 1258 / 1308 ms | **~3.7×** |
| `pretest` (type-check suite) | 4734 / 4835 ms | 944 / 1078 ms | **~4.7×** |

The two type-checking gates drop from ~9.5 s to ~2.3 s. To be precise about the claim: `pnpm test` runs mocha through `tsx`/esbuild and never touches the compiler, so **the mocha run itself is not faster** — the win is `build` and `pretest`.

`@typescript-eslint` cannot run on TypeScript 7. Every published version, canary included, declares `typescript >=4.8.4 <6.1.0` and throws at module init, which would take out `lint`, `posttest` and the pre-commit hook ([#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940)). So the two compilers sit side by side in Microsoft's documented arrangement: `tsc` is TypeScript 7 and builds; bare `typescript` is the TypeScript 6 shim that only the linter loads. Naming them this way, rather than aliasing TypeScript 7, keeps the compiler each script gets written down and survives `pnpm update --latest`. **This is temporary** — `.ai/instructions.md` records the removal steps and the two traps while it stands.

Emitted output was diffed between compilers: 432 files, 81 differ, all cosmetic — sourcemaps, temp-variable naming, declaration ordering, one quote-style change. No semantic change to shipped runtime behaviour.

## Notable

- **A latent bug surfaced and is fixed** (`1be71d9`). `SdkQuickstartAction` built its `ValidationService` from `this.configDir` in a property initialiser, but `configDir` is a constructor parameter property. It only worked because downlevel emit assigns parameter properties first; under the native class-field semantics an ES2024 target selects, `this.configDir` is undefined there. Worth reviewing on its own.
- **`target` moves es2017 → es2024**, which also selects native class-field semantics. That is a behavioural change beyond syntax; the suite (480 passing) is the evidence. `useDefineForClassFields: false` is the one-line fallback if you would rather split it out.
- **The portal-only Node gate is gone.** It existed for a TanStack Start build requirement, not a CLI-wide one; two gates holding the same number drift. `engines` is only advisory, so the floor is enforced at run time by an `init` hook, with a test asserting the constant matches `engines.node`.
- **Releasing on Node 24 removed the `npm install -g npm@latest` step** — Node 22 bundles npm 10.9.8, below the 11.5.1 OIDC publishing needs; Node 24 bundles 11.19.0. The constraint is now a comment on the matrix, because pinning that job back would break publishing silently.
- **`@types/node` is pinned to 24.13.4, not 24.13.6** — newer releases are inside the repo's own 7-day `minimumReleaseAge` window, and bypassing a supply-chain defence for a types bump is not a trade worth making.

## Deviating from what was agreed

The standalone pnpm in `test.yml` is **kept**, not reverted to corepack. Its original reason (pnpm 11 needs Node 22.13, the floor was 22.12) is gone, but it is a SHA-pinned executable needing no Node of its own, where corepack resolves pnpm from the registry at run time and needs the three-step `setup-node` dance. Reverting would have cost supply-chain hygiene in the most security-relevant workflow to remove one step. The comment now states the reason that actually holds. Happy to do the full revert if you disagree.

## Verification

`pnpm install --frozen-lockfile`, `pnpm build`, `pretest`, **480 passing / 14 pending / 0 failing**, lint clean, built CLI runs, and the engine guard was exercised on both paths.

Lint was run scoped rather than via `pnpm test`: `posttest` is `eslint .`, which also sweeps a stale `.claude/worktrees` copy in my working tree. CI checkouts do not have that, so CI is unaffected — but it does mean `pnpm test` fails locally for anyone with a worktree. Not fixed here; say the word and I will add `.claude` to the eslint ignores.

BREAKING CHANGE: `engines.node` is now `>=24.0.0`. Node 22 is no longer supported.


.github/workflows/check_build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '22', '24' ]
+        node: [ '24', '26' ]
```
saeedjamshaid, line 19: **Blocking** — this job still runs `corepack enable` (line 37), and Corepack is no longer distributed with Node.js 25 and newer.

On the Node 26 leg, `actions/setup-node` puts a Node 26 toolcache bin dir first on PATH. That dir contains no `corepack`, so `corepack enable` either fails with `corepack: command not found` or silently picks up the runner image's preinstalled Node's corepack. Either way the following `cache: 'pnpm'` step and `pnpm install --frozen-lockfile` run with no pinned pnpm, and the job breaks.

`test.yml` was deliberately moved off corepack to a SHA-pinned standalone pnpm for exactly this class of problem — `check_build.yml` and `release.yml` were left behind. Suggest lifting the same `pnpm/action-setup` block (`version: 11.0.8`, `standalone: true`) into both.

Refs: [Node.js TSC votes to stop distributing Corepack](https://socket.dev/blog/node-js-tsc-votes-to-stop-distributing-corepack) · [nodejs/nodejs.org#7555](https://github.com/nodejs/nodejs.org/issues/7555)

.github/workflows/release.yml
```diff
@@ -15,7 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '22' ]
+        # Keep this at 24 or newer: npm trusted publishing (OIDC) needs npm >= 11.5.1,
+        # and Node 24 is the first release whose bundled npm clears that.
```
saeedjamshaid: **Blocking** — the comment is factually wrong, and it is the only thing left guarding OIDC publishing after the `npm install -g npm@latest` step was removed below.

Node 24.0.0 through 24.4.x bundle **npm 11.3.0**, which is below the 11.5.1 that trusted publishing needs. npm 11.5.1 first ships in **Node 24.5.0**.

This works today only because `'24'` resolves to the newest 24.x. But the comment tells future maintainers "Node 24 is the first release whose bundled npm clears that" and "keep this at 24 or newer" — so pinning to `24.0`, `24.4`, or any environment resolving an early 24.x drops `semantic-release` off OIDC silently, which is the exact failure the removed step existed to prevent.

Suggest pinning the matrix to `'24.5'` (or higher) and correcting the comment to name 24.5.0 as the floor.

.github/workflows/release.yml
```diff
@@ -50,11 +52,6 @@ jobs:
         working-directory: 'cli'
         run: pnpm build
 
-      # npm trusted publishing (OIDC) requires npm >= 11.5.1; the runner's bundled
-      # npm (shipped with Node 22) is older, so upgrade it before publishing.
-      - name: Upgrade npm for OIDC trusted publishing
-        run: npm install -g npm@latest
-
       - name: Release
```
saeedjamshaid, line 54: **Blocking** (same root cause as `check_build.yml`) — this workflow still bootstraps pnpm through `corepack enable` at line 38.

Two problems:

1. Corepack is absent from Node 25+. Node 26 is already in this PR's test matrix, so the next time the release floor moves, the release job dies at `corepack enable` and cannot publish.
2. Unlike `test.yml`'s SHA-pinned standalone pnpm, corepack downloads pnpm@11.0.8 from the registry on **every release run**. The PR's own rationale for standalone pnpm — supply-chain hygiene in the most security-relevant workflow — applies more strongly to the workflow that actually publishes than to the one that runs tests.

.github/workflows/test.yml
```diff
@@ -14,16 +14,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node: ['22.12', '24']
+        node: ['24', '26']
```
saeedjamshaid, line 17: **Worth checking before merge** — job names interpolate the matrix (`${{ matrix.os }} / Node ${{ matrix.node }}` here, `Node ${{ matrix.node }} sample` in `check_build.yml`), so renaming the values renames every status check these workflows report.

Checks previously reported as `ubuntu-latest / Node 22.12` and `Node 22 sample` will never report again. If `dev` or `main` branch protection lists those names as required, every subsequent PR hangs on checks that can no longer run — and the fix is in repo settings, invisible from this diff.

Not a code change; just needs the protection rules updated in the same change window.

src/hooks/node-version.ts
```diff
@@ -0,0 +1,19 @@
+import { Hook } from '@oclif/core';
+
+// Kept in step with `engines.node` in package.json. npm and pnpm only warn when the
+// installed Node is older, so without this the install succeeds and the failure surfaces
+// later as an unrelated-looking crash.
+export const MINIMUM_NODE_MAJOR = 24;
```
saeedjamshaid: **Should fix** — the floor is mirrored here as a hardcoded constant, and `test/hooks/node-version.test.ts` exists solely to assert the mirror. oclif has already parsed the manifest by the time an init hook fires:

```ts
const declared = this.config.pjson.engines?.node;
```

Deriving the major from that makes `package.json` the single source of truth and lets you delete both `MINIMUM_NODE_MAJOR` and the whole test file. As written, every floor change is a three-place edit (package.json, this constant, README) guarded by a test that only re-checks two of them.

src/hooks/node-version.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const hook: Hook.Init = async function () {
+  const major = Number(process.versions.node.split('.')[0]);
+  if (major < MINIMUM_NODE_MAJOR) {
```
saeedjamshaid: **Should fix** — this compares only the major, while `engines.node` is a full semver range.

Concretely: if you take the `release.yml` finding and move `engines.node` to `>=24.5.0` (a realistic move, since npm 11.5.1 needs Node 24.5.0), then `engines.node.replace(/^\D*/, '').split('.')[0]` still yields `24`, the mirror test still passes, and this hook still waves through Node 24.0.0 — a runtime the manifest says is unsupported.

The test's title, "requires the same major that package.json declares as the engine floor", reads as a stronger guarantee than it gives. Comparing with a real semver satisfies check against the declared range would close both gaps.

test/hooks/node-version.test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect } from 'chai';
+import fs from 'node:fs';
+import { MINIMUM_NODE_MAJOR } from '../../src/hooks/node-version';
+
+describe('node version hook', () => {
+  it('requires the same major that package.json declares as the engine floor', () => {
```
saeedjamshaid: **Should fix** — this is the only test for the new hook, and it asserts a constant equals `package.json`. Nothing exercises the hook's behaviour.

The deleted `MINIMUM_NODE_VERSION` check in `PortalProjectService.runtimeProblem()` *was* covered — by "stops before anything else on a Node this build does not support" in both `generate.test.ts` and `serve.test.ts`. Those two assertions were repurposed to the vite-missing branch in this PR, so that coverage is gone rather than moved.

Every one of these ships with a green suite today:
- invert the comparison to `major > MINIMUM_NODE_MAJOR`
- read `process.version` instead of `process.versions.node` (`Number('v24')` is `NaN`, so the comparison is always false)
- drop `{ exit: 1 }` so oclif swallows the failure

A test that runs the hook with a stubbed version on both sides of the floor would cover it.

src/infrastructure/portal-project-service.ts
```diff
@@ -51,16 +48,8 @@ export class PortalProjectService {
   private readonly fileService = new FileService();
   private readonly require = createRequire(import.meta.url);
 
-  /** Checks this Node build can run the portal build at all, before any work is done. */
+  /** Checks this installation can run the portal build at all, before any work is done. */
   public runtimeProblem(): string | null {
```
saeedjamshaid, line 52: **Should fix** — deleting `MINIMUM_NODE_VERSION` removes the only record that TanStack Start's build has its own Node floor.

The two gates were collapsed because they held the same number, but they encode different facts: the CLI's supported-runtime policy (now the init hook) and what the bundled Vite/TanStack toolchain can actually build with. A future template bump that needs, say, Node 26 now has nowhere to declare it, and the failure surfaces as a raw Vite or esbuild crash inside the temp project rather than as the `runtimeProblem()` message the actions are already wired to display.

A one-line constant asserted against `MINIMUM_NODE_MAJOR` would keep the fact on record without reintroducing a second runtime check.

.ai/instructions.md
```diff
@@ -37,6 +37,33 @@ pnpm apimatic <command> -i <abs-path-to-dir-containing-src>   # -i: for commands
...
+`@typescript/typescript6`, and exists only so `@typescript-eslint` can load: every published
+version of it declares `typescript >=4.8.4 <6.1.0` and throws at module init on TypeScript 7,
+which would take out `lint`, `posttest` and the pre-commit hook. TypeScript 6 is also
+available directly as `tsc6` if you need to compare the two.
+
+This is Microsoft's documented side-by-side arrangement, and it is temporary. **Delete it
+once typescript-eslint supports TypeScript 7** (tracked at
+<https://github.com/typescript-eslint/typescript-eslint/issues/10940>): drop the
+`@typescript/native` alias, point `typescript` back at `^7`, and bump `@typescript-eslint/*`
+to whatever release added support. Nothing else depends on the arrangement — no source file
```
saeedjamshaid: **Should fix** — "Nothing else depends on the arrangement — no source file imports `typescript`" is true of `src/`, but `test/e2e/portal-build.test.ts` resolves the alias directly:

```ts
require.resolve('@typescript/native/package.json')
```

Following the removal steps as recorded — drop the `@typescript/native` alias, point `typescript` back at `^7` — makes that throw `ERR_MODULE_NOT_FOUND`, so the e2e "type-checks against the packages it is built with" test fails across the whole platform matrix with an error that reads like a broken install.

Worth naming that call site here, and pointing the comment above it in the test back at this section.

.vscode/launch.json
```diff
@@ -6,51 +6,10 @@
...
-            // "--ignore=C:/Users/User/Desktop/demo/portal/content,C:/Users/User/Desktop/demo/portal/static/images/logo.png",
-            // "--no-reload",
-            // "--auth-key=\"invalidApiKey\"",
-            // "--help"
-        ],
+        "program": "${workspaceFolder}/bin/run.js",
+        "args": [],
         "console": "externalTerminal",
-        "preLaunchTask": "tsc: build - tsconfig.json"
+        "preLaunchTask": "npm: build"
```
saeedjamshaid: **Should fix** — `"npm: build"` won't resolve in this repo, so the repaired config still won't launch.

VS Code's script task provider labels tasks by the package manager it detects. With `pnpm-lock.yaml` present and `"packageManager": "pnpm@11.0.8"` in `package.json`, the task is contributed as **`pnpm: build`**. Pressing F5 on "Debug APIMatic CLI" pops "Could not find the task npm: build" and never runs — the same class of breakage this commit set out to fix.

```suggestion
        "preLaunchTask": "pnpm: build"
```

.gitignore
```diff
@@ -177,7 +177,6 @@ $RECYCLE.BIN/
 # directories such as portal-template/src/lib.
 /lib/
 tmp/
```
saeedjamshaid, line 179: **Nit / intent** — removing the `.vscode/launch.json` ignore (the line that was here) makes a file whose entire history is personal scratch configuration into a tracked, shared file.

The same commit deletes 44 lines of one developer's commented-out local paths (`C:/Users/User/Desktop/...`) from that file — which is what the ignore existed to keep out of the repo. Now any developer who edits `args` to debug a command either carries a permanently dirty working tree or commits their local paths back in.

Tracking a sanitised shared config is defensible, but the ignore removal and the cleanup pull in opposite directions and nothing in the PR records which outcome was intended. A line in the commit body would settle it.

saeedjamshaid (commented):
## Review: Node 24 + TypeScript 7 migration

13 findings, 12 of them left inline. The CI ones are the real blockers — the Node 26 leg of `check_build` cannot work as written, and `release.yml` is a ticking version of the same bug.

**Blocking**
1. `check_build.yml` bumped to Node 26 but still bootstraps pnpm with `corepack enable` — Corepack is no longer distributed with Node 25+.
2. `release.yml` has the same corepack bootstrap, plus it pulls pnpm from the registry at release time.
3. The comment justifying the removal of `npm install -g npm@latest` is factually wrong: the OIDC floor is Node **24.5**, not 24.0.

**Worth checking before merge**
4. Renaming the matrix values renames every reported status check (`ubuntu-latest / Node 22.12`, `Node 22 sample`). If branch protection on `dev`/`main` requires those names, subsequent PRs will hang on checks that can no longer report — and the fix is in repo settings, not this diff.

**One finding that has no line in the diff:**

`test/mocha.opts` survived the `ts-node` removal. Its first line is `--require ts-node/register`, naming a package that no longer installs, and its `--timeout 5000` contradicts the real `--timeout 99999`. The file is already dead (mocha dropped `mocha.opts` support in v8; this repo is on mocha 11 and passes flags on the command line), but it should have been deleted in the same commit that dropped `ts-node` and the `"ts-node": { "esm": true }` block from `tsconfig.json`.

The rest are inline, labelled **Blocking** / **Should fix** / **Nit**.

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

test/actions/portal/generate.test.ts
```diff
@@ -92,13 +92,13 @@ describe('GenerateAction', () => {
...
-  it('stops before anything else on a Node this build does not support', async () => {
-    runtimeProblem.returns('Building a portal needs Node 22.12.0 or newer.');
+  it('stops before anything else when the installation cannot build a portal', async () => {
+    runtimeProblem.returns("The portal build dependency 'vite' is missing from this installation.");
 
     const result = await execute();
 
     expect(result.isFailed()).to.be.true;
-    expect(prompts.runtimeUnsupported.calledOnceWith('Building a portal needs Node 22.12.0 or newer.')).to.be.true;
+    expect(prompts.runtimeUnsupported.calledOnceWith("The portal build dependency 'vite' is missing from this installation.")).to.be.true;
```
saeedjamshaid: **Nit** — this line is 138 characters, over the repo's Prettier `printWidth: 120`. The same 138-char line is at `test/actions/portal/serve.test.ts:87`.

`.prettierrc` sets `printWidth: 120` and the lint-staged entry runs `prettier --write` on `*.{js,ts,tsx}`, which covers test files. Nothing fails today only because eslint carries no width rule and the git hooks are not reliably installed — but the next contributor to touch either file with working hooks gets an unrelated reformat mixed into their diff.

saeedjamshaid:
## Second pass: is the change justified, and what are the risks?

Following up on [the line-level review](https://github.com/apimatic/apimatic-cli/pull/347#pullrequestreview-5287016119) with a look at the premise rather than the diff. I verified the load-bearing claims rather than taking the description at its word; where I could not, I say so.

### Is it justified?

**The Node 24 floor: yes, clearly.** The reasoning holds up — Node 22 is EOL 2027-04-30, inside the supported life of the major being cut. And this is not a major invented to carry the change: `release.config.cjs:12-14` already lists a `1.x` branch "once 2.0.0 prereleases exist", so a v2 was planned independently. A floor bump is exactly what a major is for.

**TypeScript 7: justified, but it is the weaker half and it carries all the complexity.** Two things worth putting on the record, both checked against the registry today:

- **TS 7.0.2 is the `latest` dist-tag, published 2026-07-08** — GA, roughly 2.5 months old, not a preview. That is a much better footing than "native compiler rewrite" suggests on its face.
- **`@typescript-eslint/parser@8.70.1`, the current latest, still declares `typescript >=4.8.4 <6.1.0`.** So the dual-compiler arrangement is genuinely *forced*, not a stylistic preference. The PR's account of [#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940) is accurate.

The framing I would add: these two changes have very different profiles. The floor is high-justification, low-risk and time-sensitive. TypeScript 7 buys roughly 7 s per build/type-check cycle — real developer experience, but no security or user-facing benefit — at the cost of carrying an inverted-name two-compiler setup for an *unbounded* period, gated on a third party with no ship date. Bundled, the TypeScript 7 decision inherits the floor's urgency.

The counterargument is decent: `es2017 → es2024` is itself a semantics change that belongs in a major, and doing it once beats doing it twice. My read is **keep it, but decide deliberately** — `d6ff5d8c` is cleanly separable if the team would rather not carry the arrangement yet.

### Risks

**1. Class-field semantics — verified clean.** This is the scariest part of the PR, so rather than trust the suite I walked the AST of both branches looking for eager field initialisers that read a constructor parameter property:

- `dev`: exactly one hit — `src/actions/sdk/quickstart.ts:29` — precisely the bug `adb9a3cd` fixes.
- this branch: **zero**.

The five `execute` arrow-function fields are lazy (the body evaluates at call time, long after the constructor), so they are safe. The three `language` fields in the event subclasses are constructor-assigned with no base-class accessor to shadow. **The claim checks out, and there are no other instances of the pattern.** Scope of that check: `src/` only, that pattern plus declared-but-unassigned subclass fields. The repo uses no decorators, so the third `useDefineForClassFields` hazard does not apply.

**2. The `1.x` maintenance branch does not exist yet.** `git ls-remote --heads origin` returns only `beta`, `dev` and `main`. Published `latest` is **1.5.0**. `release.config.cjs` and `release.yml` both already reference `1.x`, but once `main` advances to 2.0.0 there is no correct commit left to branch v1 maintenance from without surgery. `main` is still on v1 today, so **cut `1.x` from `main` before this progresses toward a release branch.** This is the most actionable risk outside the CI blockers.

**3. Nothing publishes on merge — there is room to sequence.** The base is `dev`, which is not in `release.yml`'s trigger list (`alpha, beta, main, 1.x`). The 2.0.0 cut happens later, when `dev` reaches a release branch. That is what makes point 2 sequenceable rather than urgent.

**4. The inverted names are protected only by documentation.** `typescript` resolves to TypeScript 6; the `tsc` binary is TypeScript 7. Anyone reasoning "the typescript dependency is the compiler" is wrong, and `pnpm outdated` lists both under the real name. Worth a CI assertion — `tsc --version` starts with 7, the `typescript` package resolves to 6 — so that a bad `pnpm update` fails loudly instead of silently swapping compilers. Documentation is the right place to explain the arrangement, but the wrong place to enforce it.

**5. The emit diff is the load-bearing claim, and it is not reproducible from the PR.** "432 files, 81 differ, all cosmetic" is the entire basis for *no semantic change to shipped runtime behaviour*, and it rests on one local run recorded in prose. I could not reproduce it. If that claim matters — and for a compiler swap that emits the published artefact, it should — it belongs in CI or attached as an artefact rather than as a sentence in the description.

**6. Declaration emit reaches consumers.** `tsconfig.json` sets `declaration: true` and the package ships `types: lib/index.d.ts`. Almost certainly vestigial oclif scaffolding for a `bin`-only package — but if anything imports `@apimatic/cli` as a library, TypeScript 7's declaration emit is in the blast radius. Worth confirming whether `types` should simply be dropped.

**7. The floor lands on merge, not at release.** The new init hook hard-exits on anything below Node 24, so any contributor still on Node 22 or 23 is blocked from running the built CLI from the day this merges — well before 2.0.0 ships. Intended behaviour, but it is worth an explicit heads-up to the team rather than letting people discover it as a crash.

**8. `.claude` is missing from the eslint ignores — confirmed.** `eslint.config.js:52-62` ignores `lib`, `node_modules`, `portal-template`, `tmp` and the `test-*` directories, but not `.claude`. Claude Code creates worktrees under `.claude/worktrees` by default, so `posttest` (`eslint .`) will sweep them and `pnpm test` will fail locally for anyone who uses one. The PR offers the one-line fix — worth taking.

**9. Scope.** Nine commits spanning engine floor, compiler swap, a latent runtime bug fix, CI matrix, VS Code config, `.gitignore`, mock-fs removal and docs. The mock-fs removal is a genuine prerequisite (Node 26.8's `fs` rewrite breaks its binding interception, and CI now tests 26), not creep. But `adb9a3cd` is a real runtime bug fix that deserves its own review attention — exactly the kind of change that gets waved through inside a large build PR. The description flags it, which is the right instinct.

### Bottom line

Justified, and the engineering is better than the average migration PR. The description is unusually self-critical, and the one claim verifiable end-to-end — the class-field hazard — held up exactly. The blockers are the three CI findings in the line-level review, not the premise.

- **Before merge:** fix the corepack and OIDC-floor issues, cut `1.x` from `main`, add `.claude` to the eslint ignores.
- **Before release:** get the emit-diff check into CI, or accept it explicitly as a one-time attestation.

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

.gitignore
mrafnadeem-apimatic: I'll `git rm` it and then put it in gitignore.

.vscode/launch.json
mrafnadeem-apimatic: deleted the launch.json file entirely

.github/workflows/test.yml
mrafnadeem-apimatic: checks are running fine so I guess this is not a big problem

portal-template/vite.config.ts
```diff
@@ -38,6 +38,10 @@ export default defineConfig(async () => {
       }),
       react()
     ],
+    // The prerender pass fetches every page from a preview server on `localhost`. Where that
+    // listens on `::1` alone, a connect that stalls under load falls back to 127.0.0.1 and is
+    // refused, failing the build; binding IPv4 leaves the fetch a single address to reach.
+    preview: { host: '127.0.0.1' },
```
saeedjamshaid, line 44: **Worth a second look** — this pins the *bind* to IPv4, but the *fetch target* is still `localhost`, resolved inside TanStack Start. The two ends can still disagree.

On a host where `localhost` resolves only to `::1` — IPv6-only containers being the realistic case — the prerender fetch can now never reach the server, which is the exact inverse of the failure this fixes. What saves the common case is `autoSelectFamily` (default true on Node 20+) falling back to IPv4, so in practice this is very likely net-positive. But the mismatch is still there; it is just biased the other way now.

If TanStack Start exposes the prerender origin, pinning both ends to `127.0.0.1` would make it deterministic rather than dependent on Happy Eyeballs. If it does not, worth saying so in the comment — "binding IPv4 leaves the fetch a single address to reach" reads as though the fetch is constrained too, and it is not.

Separately: the trigger described ("a connect that stalls under load") is a flake, and there is no link to the run that caught it. The e2e does have real coverage here — `test.yml:51` sets `APIMATIC_E2E: '1'` across all six cells — so a run URL in the commit body would let the next person tell whether it recurred or whether this fixed it.

test/infrastructure/portal-project-service.test.ts
```diff
@@ -44,10 +45,25 @@ describe('PortalProjectService', () => {
...
+  // floor above the CLI's must raise `engines.node` with it rather than crash inside Vite.
+  it('supports only Node versions every template dependency supports', () => {
+    const readManifest = (directory: string) =>
+      JSON.parse(fs.readFileSync(path.join(directory, 'package.json'), 'utf8'));
+    const supported = readManifest('.').engines.node;
+
+    for (const dependency of TEMPLATE_DEPENDENCIES) {
+      const required = readManifest(path.join('node_modules', dependency)).engines?.node;
+      if (required !== undefined) {
+        expect(semver.subset(supported, required), `${dependency} needs Node ${required}`).to.be.true;
```
saeedjamshaid, line 62: **Nit — worth a note in the test.** `semver.subset` against an open-ended `supported` range has a sharp edge: the moment any template dependency adopts a *capped* range (`^20 || ^22 || ^24` is a common shape), this flips to false and CI goes red.

That is arguably the test doing its job — CI tests Node 26, so a capped dependency genuinely is news. The trap is that the obvious way to make it green again is capping `engines.node` to match, which is usually the wrong call. A line saying what to do when it fires (check whether the cap is real, and prefer waiting for a dependency release over narrowing the CLI's own range) would save the next person that decision.

For the record, I resolved all sixteen entries against the installed tree and every declared range is open-ended upward today, so nothing is close to tripping it.

test/infrastructure/portal-project-service.test.ts
```diff
@@ -44,10 +45,25 @@ describe('PortalProjectService', () => {
...
+  it('reports no runtime problem with the dependencies installed', () => {
     expect(service.runtimeProblem()).to.be.null;
   });
 
+  // The init hook is the only Node version gate, so a template dependency that raises its own
+  // floor above the CLI's must raise `engines.node` with it rather than crash inside Vite.
+  it('supports only Node versions every template dependency supports', () => {
+    const readManifest = (directory: string) =>
+      JSON.parse(fs.readFileSync(path.join(directory, 'package.json'), 'utf8'));
+    const supported = readManifest('.').engines.node;
```
saeedjamshaid, line 57: **Nit** — `readManifest('.')` and `path.join('node_modules', dependency)` below are resolved against the current working directory, while every other test in this file builds absolute paths from `os.tmpdir()`. It works because mocha runs from the repo root, but it is the one test here that would break if that ever stopped being true.

There is also a subtler mismatch: this reads the top-level `node_modules/<dep>` entry, whereas `runtimeProblem()` resolves the same dependency through `createRequire` from the installed CLI. Under this pnpm layout they land on the same package, but they are different resolution mechanisms, so the test can in principle assert against a different copy than the code under test uses.

.github/workflows/release.yml
```diff
@@ -27,15 +29,14 @@ jobs:
...
-          node-version: ${{ matrix.node }}
+          version: 11.0.8
+          standalone: true
 
-      - name: enable corepack
-        run: corepack enable
-
-      - name: configure pnpm cache
+      - name: setup node
         uses: actions/setup-node@v4
```
saeedjamshaid, line 40: **Nit** — the pnpm action just above is SHA-pinned, but `actions/checkout@v4` and `actions/setup-node@v4` in this same job are still floating tags. Given the stated rationale for standalone pnpm was supply-chain hygiene specifically in the publishing workflow, the inconsistency stands out.

These are GitHub-owned actions and a `v4` tag is a much smaller exposure than a registry fetch, so this is genuinely minor — but if the pinning is worth doing here, it is worth doing to all three.

saeedjamshaid (commented):
## Re-review at `a404e3f`

All thirteen findings from the first pass are addressed, and all ten checks are green including the full six-cell matrix. Notes on what I verified rather than took on trust:

- **The engine guard now reads the manifest.** `this.config.pjson.engines?.node` + `semver.satisfies` closes both the mirrored-constant and major-only-comparison findings in one change, and the three new tests exercise the outside-range, inside-range and no-`engines` paths for real. `semver` is in `dependencies`, not `devDependencies` — correct, since the init hook imports it at runtime in the published package.
- **The template-floor test is a better answer than the constant I suggested.** Asserting `engines.node` is a `semver.subset` of every template dependency's declared range keeps the fact on record *and* checks it. I resolved all sixteen `TEMPLATE_DEPENDENCIES` against the installed tree: every range is open-ended upward (`>=20.19`, `>=22.12.0`, `^20.19.0 || >=22.12.0`, `>=20`, `>=0.10.0`), so the assertion holds across the board today.
- **Both workflows are correct.** Standalone pnpm ordered before `setup-node`, single node step, `cache: 'pnpm'` and `cache-dependency-path` retained.
- **Formatting is clean.** No line over 120 characters in any changed file.
- `test/mocha.opts` and `.vscode/launch.json` are both gone, `.gitignore` re-ignores the latter, and `.ai/instructions.md` now names the e2e call site.

Four new points on the five new commits, none of them blocking. The only substantive one is the preview-server bind.

Still outstanding outside this PR, for whoever has admin: cut `1.x` from `main` while it is still on v1, and preserve the `BREAKING CHANGE:` footer when squashing — the repo's `squash_merge_commit_message` is `COMMIT_MESSAGES` so it is there by default, but trimming it turns a `build` type into no release at all rather than 2.0.0.

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

saeedjamshaid (approved):

PR #348, opened 2026-09-22
feat(config): replace portal.json and plugin-config.json with src/apimatic.json

saeedjamshaid:
One file, `src/apimatic.json`, replaces `src/portal.json` and `src/plugin-config.json`. It holds
`portal`, `plugin` and `languages`; the portal reads the first, the plugin commands the other two,
and a malformed block one command owns never stops another. Implements
`.ai/plans/apimatic-config.md`, which follows on from #346.

**Merge only once the backend reads `plugin` and `languages` from the `apimatic.json` inside the
uploaded `src/`.** Nothing is synthesized for it and no compatibility is kept: `plugin generate`
zips `src/` as it is, and a project with an old `plugin-config.json` records its SDKs and identity
again.

**Do not merge `dev` into `main` until the series ends.** `dev` already carries a breaking marker
(#347, Node 24), so its next release is 2.0.0 whatever this PR does; the reason to wait is that a
half-finished series would ship, with `src/portal.json` gone and `quickstart` not yet adopting an
existing `apimatic.json`. This commit is an ordinary `feat:`; the `BREAKING CHANGE:` footer that
describes the config move rides on the last PR. Drafts for the release notes and the footer are in
the plan's section 13.

## The format

```json
{
  "schemaVersion": 1,
  "portal": {
    "title": "Petstore",
    "description": "...",
    "logo": "static/logo.png",
    "siteUrl": "https://docs.example.com"
  },
  "plugin": {
    "pluginId": "petstore-context",
    "pluginName": "Petstore Context",
    "pluginVersion": "1.0.0",
    "license": "MIT"
  },
  "languages": {
    "csharp": {
      "source": { "repositoryUrl": "https://github.com/acme/petstore-csharp", "branch": "main" },
      "package": { "packageId": "Petstore.Sdk", "version": "1.0.0" },
      "codegenVersion": "v3"
    }
  }
}
```

The blocks hold exactly what the two old files held: `portal` is the old `portal.json` object,
`plugin` is the identity `plugin-config.json` carried at its root, and `languages` is its
`languages` map with the entry shapes unchanged. `schemaVersion` is written on create and never
rewritten; anything other than `1` is a root finding that stops every command. `portal` belongs to
`portal generate`, `portal serve` and `quickstart`; `plugin` and `languages` belong to
`sdk publish`, `plugin generate` and `plugin publish`. Nothing else in the file is inspected, so a
block a later PR adds does not break this one.

## How it works

- `ApimaticConfigDocument` (`src/types/apimatic-config/document.ts`) parses the text once and
  reports `ConfigFinding`s tagged with the block they belong to. Only an empty file, invalid JSON
  or a non-object root fails the parse outright; a bad `plugin.pluginVersion` is a finding on
  `plugin`, and a `portal` that is not an object is a finding the portal path alone sees.
- `ApimaticConfigContext` (`src/types/apimatic-config-context.ts`) owns the file. `read()` gives
  `missing`, `unparseable` or `parsed`; `merge(blocks, apply)` refuses when the root or any listed
  block has findings, otherwise rewrites the document keeping key order, the file's own
  indentation and its trailing newline. The write goes through a new
  `FileService.replaceContents`, which writes a temporary file beside the target and renames it,
  so a crash never leaves a half-written `apimatic.json`.
- `PluginConfigContext` and `PortalSourceContext` are the two consumers. The first keeps its
  public shape; the second changes it in one place, the `invalidConfig` problem gaining
  `missingPortal`, which the shared source prompt reads. Otherwise the actions, prompts and
  commands above them changed only where a message named the old file. `PluginConfigPresent` is
  untouched.

Messages say `apimatic.json` everywhere the old names appeared, and portal findings are spelled
as paths (`'portal.title' is required and must be a non-empty string.`) so the user knows which
block to open. A file that exists without a `portal` block, the shape `sdk publish` alone leaves
behind, gets the same "run quickstart" line a missing file does; a `portal` that is present but
malformed does not, since there is a block to fix. The README's *Upgrading from 1.x* section
describes the file and says `src/plugin-config.json` is no longer read.

## Worth a reviewer's attention

- **The invariant that matters is tested at both levels.** `test/actions/plugin/record-sdk.test.ts`
  runs `sdk publish` against a file whose `portal` block would make `portal generate` refuse, and
  the language is recorded with that block left as it was. `plugin-config-context.test.ts` and
  `portal-source-context.test.ts` cover each direction at the context level.
- **Indentation is read from the first indented line.** A file whose first root key shares its
  line with the opening brace is rewritten at the nested indent it finds there. Cosmetic: the
  JSON and its key order are kept. `JSON.stringify` also caps an indent at ten characters.
- **The temporary file sits beside the target inside `src/`**, the directory `plugin generate`
  zips. A failed write removes it; only a hard kill between write and rename could leave one.
- **`ApimaticConfigContext.exists()` has no caller yet.** The quickstart PR, which adopts a
  directory that already holds the file, is what uses it.
- **The `--update-plugin-config` flag keeps its name.** Renaming it is a separate breaking change
  and the last PR of the series is the place for it.
- **`pnpm readme` is still broken by the CRLF `README.md`** (see #346). The five changed
  description lines were applied by hand and checked against a regeneration on an LF copy.
- **The plans beside this one were amended, not rewritten.** `portal-config.md`,
  `portal-navigation.md` and `fumadocs-portal.md` carry a dated note that the file moved into
  `src/` and that the `sdks.languages` idea gave way to the shared top-level `languages`.

First of a series: next, `quickstart` adopts a directory that already holds an `apimatic.json`;
then `portal generate` and `plugin generate` require a `languages` entry.

## Testing

668 unit tests pass with 19 pending. New suites cover the document (findings per block, BOM,
`schemaVersion`), the context (missing, unparseable, refusal on findings, formatting preserved
across a merge, indentation detection) and the atomic write (temporary file removed on failure).
The plugin action tests open the uploaded zip and assert `apimatic.json` is at its root and
`plugin-config.json` is absent. The portal fixture was renamed to
`test/resources/portal-inputs/default/apimatic.json`.


src/types/portal-source-context.ts
```diff
@@ -143,15 +140,52 @@ export class PortalSourceContext {
...
    */
   public async scaffold(specPath: FilePath): Promise<void> {
     await new SpecContext(this.specDirectory).install(specPath);
 
     const config = await this.suggestedConfig(specPath);
-    await this.fileService.writeContents(this.configFile, JSON.stringify(config, null, 2) + '\n');
+    // The directory is empty when quickstart runs this, so the merge always creates the file.
+    // A fault is thrown as the plain write before it was: there is no publish here to protect.
+    const written = await this.configContext.merge(['portal'], (document) => document.with('portal', config.toJSON()));
+    if (written.isErr()) {
```
MuHamza30, line 192: **merge** already returns the failure as a result, so could **scaffold** return it too, and let the quickstart action report it through its prompts and return **ActionResult.failed()**? As written, this is a new explicit throw in a context that **quickstart.ts:154** doesn't catch, which goes against the "no uncaught throws above infrastructure" rule.

src/types/plugin-config-context.ts
```diff
@@ -111,147 +103,93 @@ export class PluginConfigPresent {
...
 export class PluginConfigContext {
-  private readonly fileService = new FileService();
-
-  constructor(private readonly buildDirectory: DirectoryPath) {}
+  private readonly configContext: ApimaticConfigContext;
 
-  private get configPath(): FilePath {
-    return new FilePath(this.buildDirectory, new FileName('plugin-config.json'));
+  constructor(buildDirectory: DirectoryPath) {
+    this.configContext = new ApimaticConfigContext(buildDirectory);
```
MuHamza30: The context rules ask for constructor parameters to be declared `private readonly`. Could this be:

```ts
private readonly configContext = new ApimaticConfigContext(this.buildDirectory);

constructor(private readonly buildDirectory: DirectoryPath) {}
```

src/prompts/plugin/generate.ts
```diff
@@ -54,7 +53,8 @@ export class PluginGeneratePrompts {
 
   public pluginConfigUnreadable(reason: string, path: FilePath) {
     const message =
-      `${f.var(PLUGIN_CONFIG_FILE)} cannot be used: ${reason}. ` + `Fix or delete it at ${f.path(path)} and try again.`;
+      `${f.var(APIMATIC_CONFIG_FILE_NAME)} cannot be used: ${reason}. ` +
+      `Fix or delete it at ${f.path(path)} and try again.`;
```
MuHamza30: Now that `apimatic.json` also holds the `portal` block, "or delete it" is no longer safe advice: a user who deletes the file to fix a broken `plugin` or `languages` entry also loses their hand-written portal config (title, logo, siteUrl), and `portal generate` then fails with "No apimatic.json found".

Could we drop "or delete" here, or point at the block instead, e.g. "Fix the `plugin` block in … and try again"?

The same wording is in:
- `src/prompts/plugin/publish.ts:64`
- `src/prompts/plugin/record-metadata.ts:75`
- `src/prompts/plugin/record-sdk.ts:47`

src/types/apimatic-config/document.ts
```diff
@@ -0,0 +1,189 @@
...
+    return ApimaticConfigDocument.of({ schemaVersion: SCHEMA_VERSION });
+  }
+
+  /**
+   * Refuses only what no command could use: a file holding no JSON object at all. Everything
+   * else parses, and what is wrong inside it is carried as findings for each command to read or
+   * ignore by block, so a malformed block one command owns never stops another.
+   */
+  public static parse(text: string): Result<ApimaticConfigDocument, ConfigFinding[]> {
+    const contents = stripByteOrderMark(text);
```
MuHamza30, line 58: On `dev`, the plugin path refused a file starting with a byte-order mark (Notepad and PowerShell redirection write one). Now it's stripped here, but `plugin generate` zips `src/` as is, so the backend receives `apimatic.json` with the BOM still in it. Does the backend's parser handle a leading BOM? If not, we should either refuse it on the plugin path (as `nav.json` does, for the same reason) or rewrite the file without it before zipping.

MuHamza30 (changes requested):

src/types/portal-source-context.ts
saeedjamshaid: Agreed, and fixed in 9763e83 — but I took it further than the merge fault, because that alone would have satisfied the rule only on paper.

`scaffold` was never throw-free: `SpecContext.install` and the three `writeContents` calls after the merge propagated I/O faults on `dev` too, and nothing above catches them either — the throw escapes `PortalQuickstartAction`, `QuickstartAction` and `Quickstart.run` into oclif's handler, so `outro(result)` and the `QuickstartCompletedEvent` never fire.

So the whole body moved into a private `writeSourceTree` that `scaffold` wraps, and every fault comes back as a value:

```ts
public async scaffold(specPath: FilePath): Promise<Result<void, PortalScaffoldProblem>> {
  try {
    return await this.writeSourceTree(specPath);
  } catch (error) {
    return err({ kind: 'sourceUnwritable', reason: errorMessage(error) });
  }
}
```

The failure is a variant per message rather than a string, so the sentences stay in the prompts with the rest of the portal's wording, the way `resolve` already hands `PortalSourceProblem` to `reportSourceProblem`. `portal/quickstart.ts` reports it through `this.prompts.scaffoldFailed(problem, sourceDirectory)` and returns `ActionResult.failed()`.

Two tests cover it: an `apimatic.json` that cannot be parsed comes back as `configUnreadable`, and a rejected `writeContents` as `sourceUnwritable`. The existing scaffold test helper now `_unsafeUnwrap()`s, so a silent `err` fails loudly instead of surfacing as a confusing assertion further down.

Worth noting for the record: the `unreadable` branch is unreachable from quickstart today, since the action verifies the directory is empty before joining `src/`. The change is for the rule and for the file-service faults, which are reachable.

src/types/plugin-config-context.ts
saeedjamshaid: Fixed in 9763e83 — the convention point is right and line 122 was the only context not following it.

One thing to flag, though: the snippet as written does not compile. With `target: es2024`, `useDefineForClassFields` defaults to `true`, so a field initializer runs before the parameter property is assigned:

```
error TS2729: Property 'buildDirectory' is used before its initialization.
```

Under esbuild, which does not type-check, it is worse than a compile error — the field silently initializes with `undefined`.

So it went the way `PortalSourceContext` already does it at `portal-source-context.ts:64`, with the assignment in the constructor body:

```ts
private readonly configContext: ApimaticConfigContext;

constructor(private readonly buildDirectory: DirectoryPath) {
  this.configContext = new ApimaticConfigContext(this.buildDirectory);
}
```

The body reads `this.buildDirectory` rather than the parameter on purpose: nothing else in the class reads the field, and reading it here keeps it from being dead state that Sonar's S1068 could pick up.

src/prompts/plugin/generate.ts
saeedjamshaid: Good catch — fixed in 9763e83, in all four places.

I dropped "or delete" rather than naming the `plugin` block, because `getPluginConfigState` reports findings for `root` and `languages` as well (`plugin-config-context.ts:112` and `:135`), so "Fix the plugin block" would point at the wrong place when `schemaVersion` or `languages` is what is broken. The `reason` already names the offending field through `findingClause` — "its 'plugin.pluginId' must be…" — so the user knows what to edit either way.

They now read:

- `generate.ts` — `apimatic.json cannot be used: <reason>. Fix it at <path> and try again.`
- `publish.ts` — `… Fix it at <path>, then run 'apimatic plugin generate'.`
- `record-metadata.ts` / `record-sdk.ts` — `… Fix it and try again.`

src/types/apimatic-config/document.ts
saeedjamshaid: Confirmed, and fixed in 9763e83 without waiting on the backend answer.

The chain is exactly as you describe: `PluginGenerateAction` zips the build directory at `generate.ts:78` and `apimatic.json` is in it. The mark survives only when nothing rewrites the file first — `serialize` re-emits from the parsed root, so any write drops it — which makes a **repeat** `plugin generate` on an already-configured project the case that ships it. A first run writes metadata and is clean. No portal action zips or uploads `src/`, so that side is unaffected.

Rather than guess at the parser, the plugin path now takes the mark off the file before the zip:

```ts
const prepared = await configContext.removeByteOrderMark();
if (prepared.isErr()) {
  this.prompts.configNotPrepared(prepared.error, buildDirectory);
  return ActionResult.failed();
}
```

Only the mark is removed — the rest of the file is written back byte for byte, with no reserialization, so a hand-authored layout and its line endings survive a rewrite nobody asked for. A file without a mark is not touched at all, and the call sits after the guards, so a run that stops short of the zip (no metadata, no published SDKs) never rewrites anything.

That keeps the permissive parse section 2 decided on, so a project that a byte-order mark used to block still works, while nothing marked leaves the CLI. The plan is updated in both places.

Covered by a test that asserts against the actual uploaded zip, not just the file on disk, plus one that fails the command rather than uploading when the rewrite cannot be made.

MuHamza30 (approved):

PR #352, opened 2026-09-23
feat: code samples support

mrafnadeem-apimatic:


src/types/portal/openapi-document.ts
```diff
@@ -62,6 +108,56 @@ export class OpenApiDocument {
     const description = oneLine(fields.description);
     return PortalConfig.create(title, description === null ? null : cap(description, DESCRIPTION_LIMIT));
   }
+
+  private paths(): JsonObject {
+    return isObject(this.document.paths) ? this.document.paths : {};
+  }
+}
+
+function isInlinePathItem(value: unknown): value is JsonObject {
```
saeedjamshaid: **`$ref` path items get no samples, and are reported as unplaced.** `endpoints()` and `withCodeSamples()` skip any path item that is a `$ref`. Since #354, though, the portal inlines those and renders their pages.

Repro: `paths: { /health: { $ref: ./paths/health.yaml } }`, plus a catalog entry for `GET /health`. The CLI warns `No operation in 'spec' takes the code samples for GET /health`, and the built `getHealth` page has no SDK tabs.

portal-template/src/lib/request-examples.ts
```diff
@@ -0,0 +1,67 @@
...
+  private constructor(
+    public readonly location: Location,
+    private readonly examples: Map<string, Example>
+  ) {}
+
+  public static listIn(operation: unknown, pathItem: unknown): Parameter[] {
+    return [...parametersOf(operation), ...parametersOf(pathItem)].flatMap((entry) => Parameter.from(entry) ?? []);
+  }
+
+  private static from(entry: unknown): Parameter | undefined {
```
saeedjamshaid: **Parameters written as `$ref` are dropped, so the selector and the samples break.** `Parameter.from` reads `in`, `name` and `examples` straight off the node without dereferencing it. Fumadocs' own `getExampleRequests` runs `dereferenceShallow` over the same list.

Repro: `parameters: [{ $ref: '#/components/parameters/View' }]`, where `View` has `minimal` and `full` examples, and a catalog keyed `minimal`/`full`. The page has no selector, and the TypeScript tab shows "No TypeScript sample for this example." A ref to another file (`./params.yaml#/Size`, which the bundler rewrites to `#/x-ext/...`) does the same. The same parameter written inline works.

`isExampleEntry` looks like it has the same problem for an example that is a `$ref` to `components/examples`, but I didn't build that case.

portal-template/src/components/example-layout.tsx
```diff
@@ -0,0 +1,87 @@
...
+function ExampleLayout({ usageTabs, responseTabs }: Readonly<{ usageTabs: ReactNode; responseTabs: ReactNode }>) {
+  const { schema } = useRenderContext();
+  const { route, examples: bodyExamples, setExample } = useOperationContext();
+  const pathItem = schema.resolve(schema.dereferenced.paths?.[route]);
+  const operation = pathItem?.[bodyExamples[0].data.method];
+  const examples = useMemo(
+    () => requestExamples(bodyExamples, Parameter.listIn(operation, pathItem)),
+    [bodyExamples, operation, pathItem]
+  );
+  const [selectedId, setSelectedId] = useState(examples[0].id);
```
saeedjamshaid: **`x-selectedCodeSample` and `x-exclusiveCodeSample` are ignored, so the selector and the playground disagree.** Fumadocs' `OperationProvider` starts at `x-exclusiveCodeSample ?? x-selectedCodeSample`, and passes `slots.selector` as `null` for an exclusive sample. This layout starts at `examples[0]`, and always renders its own selector.

Repro: body examples `minimal` and `full`, with `x-selectedCodeSample: full`. The selector and the TypeScript tab open on "Minimal", while the playground body opens with the `full` values. With `x-exclusiveCodeSample: full`, the selector still shows.

It could start from `useOperationContext().example` when that id is in `examples`, and hide the selector when `slots.selector` is `null`.

src/types/portal/openapi-document.ts
```diff
@@ -62,6 +108,56 @@ export class OpenApiDocument {
...
+function pathItemWithSamples(path: string, pathItem: JsonObject, codeSamples: CodeSamples): JsonObject {
+  return Object.fromEntries(
+    Object.entries(pathItem).map(([key, value]) => [
+      key,
+      isOperation(key, value) ? withSamples(value, codeSamples.samplesFor(new Endpoint(key, path))) : value
+    ])
+  );
+}
+
+function withSamples(operation: JsonObject, samples: CodeSample[]): JsonObject {
```
saeedjamshaid: **Samples already in the spec survive when the catalog has none for that operation.** `withSamples` returns the operation unchanged when `samples` is empty. So an existing `x-apimatic-codeSamples`, hand-written or left from an earlier annotated spec, keeps rendering. The design doc says the merge replaces "any the operation already carries".

Repro: an operation with `x-apimatic-codeSamples: [{ lang: typescript, label: TypeScript, sources: { _default: OLD } }]` and no catalog entry. The built page shows `OLD` in the TypeScript tab.

If replacing is the intent, drop the key when there are no samples.

src/types/portal/openapi-document.ts
```diff
@@ -27,15 +41,47 @@ export class OpenApiDocument {
       const text = stripByteOrderMark(contents);
       const document: unknown = fileName.hasExtension('.json') ? JSON.parse(text) : parseYaml(text);
```
saeedjamshaid: **The CLI parses YAML differently from the bundler, so it can misread a spec that the portal renders fine.** This uses the `yaml` defaults, while json-magic's `parseYaml` plugin passes `{ merge: true, maxAliasCount: 10000 }`. Two cases that reproduce:

- A path item written as `/merged: { <<: *commonOps }` only has a `<<` key here. It gets no samples, and the CLI warns `No operation in 'spec' takes the code samples for GET /merged`, yet the portal renders the page.
- A referenced file in a subfolder of `spec/` with more than 100 aliases fails `OpenApiDocument.parse` in `refersOutside`, so its `../` refs are never followed. The copy is used anyway, and the build fails with `[OpenAPI] Failed to resolve $ref in .../tmp-.../spec/api.yaml: ../../shared/extra.yaml`. The same fixture with 50 aliases gets the "left off" warning, and builds.

Passing the bundler's options here would make the two agree.

portal-template/src/components/api-page.tsx
```diff
@@ -1,11 +1,11 @@
...
-// Only curl is generated for the request samples. Samples in other languages come from the
-// `x-codeSamples` extension on each operation in the specification, which the same tabs
-// render alongside the generated one.
-const codeUsages = createCodeUsageGeneratorRegistry();
-codeUsages.add('curl', curl);
-
-export const OpenAPIPage = createOpenAPIPage({ codeUsages, generateTypeScriptDefinitions: false });
+export const OpenAPIPage = createOpenAPIPage({
+  // Without a registry of its own, Fumadocs registers and bundles every request generator it has.
+  codeUsages: createCodeUsageGeneratorRegistry(),
```
saeedjamshaid: **Until the endpoint exists, every portal loses its request snippet.** Dropping cURL follows D19, but for now samples only come from `APIMATIC_CODE_SAMPLES_PATH`. So any `portal generate` without that variable produces operation pages with response tabs and nothing else. The example selector still shows on pages that have body examples, but the only thing it changes is the collapsed playground.

Repro: `test/resources/portal-inputs/code-samples` without the env var. `createPayment` shows the selector and the `201` response tab, and no request sample.

Could cURL stay until the endpoint ships, or at least on operations that have no samples?

src/infrastructure/portal-project-service.ts
```diff
@@ -85,6 +93,27 @@ export class PortalProjectService {
...
+    source: PortalSource,
+    codeSamples: CodeSamples
+  ): Promise<SampledSource> {
+    const specDirectory = projectDirectory.join('spec');
+    await this.fileService.copyDirectoryContents(source.specDirectory, specDirectory);
+
+    const refersOutside = await Promise.all(
+      source.specs.map((spec) => this.refersOutside(spec, source.specDirectory))
+    );
+    const unsampled = source.specs.filter((_, index) => refersOutside[index]);
```
saeedjamshaid: **The "left off" warning fires even when there are no samples.** `refersOutside` runs whatever the catalog holds, and today the catalog is empty by default.

Repro: the `code-samples` fixture, with `./schemas/error.yaml` moved to `../shared/error.yaml`, and no env var. Every `generate` and `serve` prints `Code samples are left off the reference pages of 'payments.yaml', since their $refs reach outside 'spec'.`

Skipping the copy and the check when the catalog is empty would fix this.

src/actions/portal/serve.ts
```diff
@@ -63,13 +65,22 @@ export class PortalServeAction {
...
+    this.prompts.unplacedSamples(codeSamples.value.unplacedIn(source.value.specs.map((spec) => spec.document)));
+
     const servePort = await this.networkService.getServerPort([port, 3000, 3001, 3002]);
     if (servePort !== port) {
       this.prompts.usingFallbackPort(port, servePort);
     }
 
     return await withBuildDirectory(sourceDirectory, async (tempDirectory) => {
-      const project = await this.projectService.prepare(tempDirectory, source.value);
+      const sampled = await this.projectService.addCodeSamples(tempDirectory, source.value, codeSamples.value);
```
saeedjamshaid: **`portal serve` no longer picks up spec edits.** The preview now reads the copy in the temp project, which is taken once at startup. A content edit still re-runs `source.server.ts` and re-bundles the spec, but it re-bundles the copy. So the user's edits never show until a restart. The "Live preview" note only says that changing *which* documents are in `spec/` needs a restart.

Repro: while `serve` runs, change an operation's `summary` in `src/spec/payments.yaml`, edit `content/index.md`, and reload. The old title stays. Changing the same summary in the temp copy, then editing the page again, does show it.

src/infrastructure/services/portal-artifacts-service.ts
```diff
@@ -0,0 +1,42 @@
...
+  // Stands in for a call to `/api/portal-artifacts` until that endpoint exists.
+  public async generate(): Promise<Result<CodeSamples, ServiceError>> {
+    const samplesPath = process.env.APIMATIC_CODE_SAMPLES_PATH;
+    if (!samplesPath) {
+      return ok(new CodeSamples([]));
+    }
+
+    const samplesFile = FilePath.create(path.resolve(samplesPath));
+    if (!samplesFile || !(await this.fileService.fileExists(samplesFile))) {
+      return err(ServiceError.NotFound);
```
saeedjamshaid: **Problems with the samples file don't point at the file.** A wrong path prints `Resource not found.`. Malformed JSON, or a top-level key that isn't a language (for example `"version": 1`), prints `An unexpected error occurred, please try again later. If the problem persists, please reach out to our team at 'support@apimatic.io'`. None of them mentions `APIMATIC_CODE_SAMPLES_PATH` or the path it read.

This code is temporary, but people trying the feature will hit exactly these errors. A message naming the variable and the file would help. Unknown keys could also be skipped rather than failing the build.

src/infrastructure/portal-project-service.ts
```diff
@@ -122,6 +151,35 @@ export class PortalProjectService {
...
+    }
+    return false;
+  }
+
+  private async writeWithCodeSamples(
+    spec: PortalSpec,
+    directory: DirectoryPath,
+    codeSamples: CodeSamples
+  ): Promise<PortalSpec> {
+    const file = spec.file.replaceDirectory(directory);
```
saeedjamshaid: **Build errors now name a temp file that is gone once the command exits.** The spec's path in `portal.config.json` becomes the copy, so bundler errors point at the copy.

Repro: change `./schemas/error.yaml` to `./schemas/missing.yaml` in the fixture. The build log says `[OpenAPI] Failed to resolve $ref in C:/Users/.../Temp/tmp-52760-.../spec/payments.yaml: ./schemas/missing.yaml`, and that directory is deleted when the command exits.

saeedjamshaid (commented):
I built portals from this branch (`b8d9a02`) with the real CLI, and checked the operation pages in headless Chrome. Every inline comment below reproduces, and each one says how.

**Worth fixing before merge:** `$ref` path items get no samples; `$ref` parameters break the selector; `x-selectedCodeSample` is ignored; `portal serve` stops picking up spec edits; and builds without the env var lose their only request snippet.

**One suggestion that covers most of the CLI-side comments.** The `$ref` path items, the YAML parsing mismatch, the outside-`$ref` warning, `serve` reading a stale copy, temp paths in errors, the locked-file crash and Tailwind scanning the copy all come from merging the samples by copying and rewriting `spec/` on disk. Instead, the CLI could write the catalog into the build project (next to `portal.config.json`), and `bundleSpecification` could attach `x-apimatic-codeSamples` after `inlinePathItems`. By then every path item is inline and every referenced file is embedded. That removes the copy, `refersOutside`, the unsampled-spec warning and `serialize()`, and specs whose `$ref`s leave `spec/` would get samples too. The unplaced-samples warning would then need the endpoint list from the bundled document, not the raw one. The two template comments (`$ref` parameters, `x-selectedCodeSample`) are separate, and apply either way.

src/infrastructure/portal-project-service.ts
```diff
@@ -85,6 +93,27 @@ export class PortalProjectService {
     });
   }
 
+  // A spec whose `$ref`s leave `spec/` keeps its original file: they would not resolve from the copy.
+  public async addCodeSamples(
+    projectDirectory: DirectoryPath,
+    source: PortalSource,
+    codeSamples: CodeSamples
+  ): Promise<SampledSource> {
```
saeedjamshaid: **File errors here escape as uncaught exceptions.** `addCodeSamples` returns a bare `Promise`, so a failed copy, read or write rejects straight through `GenerateAction` and `PortalServeAction`. `.ai/instructions.md` says: "Services return `Result<T, ServiceError>` ... No uncaught throws above infrastructure."

Repro: hold a lock on a file under `spec/` that no spec references (I used `spec/schemas/notes.yaml`), and run `portal generate`. It prints a raw `Error: EBUSY: resource busy or locked, copyfile ...` instead of the CLI's own failure output. On `dev` that file is never read, and the build succeeds.

src/infrastructure/portal-project-service.ts
```diff
@@ -85,6 +93,27 @@ export class PortalProjectService {
...
+    const specDirectory = projectDirectory.join('spec');
```
saeedjamshaid: Minor: the copy sits in `<project>/spec`, inside the Vite root, where Tailwind's automatic source detection scans it (`app.css` only excludes `../../dist`). Words in descriptions that happen to match utilities end up in the CSS. A description containing `skew-y-12` and `hue-rotate-180` added both rules to the built stylesheet. It's small, but `@source not '../../spec'`, or putting the copy outside the root, would avoid it.

src/types/portal/code-samples.ts
```diff
@@ -0,0 +1,73 @@
...
+
+function toCodeSample(language: Language, sources: Sources): CodeSample {
+  return { lang: language, label: languageLabel(language), sources };
+}
+
+function languageLabel(language: Language): string {
+  return LANGUAGE_CHOICES.find((choice) => choice.value === language)?.label ?? language;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
saeedjamshaid: Nit: this is the third object guard on the CLI side, next to `isObject` in `openapi-document.ts` and the existing `isJsonObject` in `src/types/apimatic-config/document.ts`. The template has its own copies in `lib/code-samples.ts`, `lib/request-examples.ts` and `openapi-bundle.server.ts`. One shared helper on each side would do. The template can't import from the CLI, so the two sides do need their own.

test/infrastructure/services/portal-artifacts-service.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
+  const withSamples = (contents: string) => {
+    const file = path.join(root, 'code-samples.json');
+    fs.writeFileSync(file, contents);
+    process.env.APIMATIC_CODE_SAMPLES_PATH = path.relative(process.cwd(), file);
+  };
+
+  it('returns no samples when no file is configured', async () => {
```
saeedjamshaid: Nit: this test reads the real environment. For anyone who has exported `APIMATIC_CODE_SAMPLES_PATH` while trying the feature, it fails with `expected [ ... ] to be empty`. Saving and clearing the variable in a `beforeEach`, then restoring it afterwards, would keep the test isolated.

.ai/plans/fumadocs-portal.md
```diff
@@ -147,8 +147,8 @@ from Google Fonts at runtime, Fumadocs page actions) except where noted.
   pages x document size: 150 Stripe operations produced 812 MB.
 - Request samples: only curl is generated (`createOpenAPIPage({ codeUsages })` with a
```
saeedjamshaid: Nit: this still says that only curl is generated, and that the SDK samples render "beside the generated one". This PR removes the curl tab.

saeedjamshaid (commented):
I re-ran the first round's fixtures against `c20bf0f`, and all 15 comments are fixed. Moving the merge into the template cleared most of them at once. The unit tests and lint pass. Four new points came up, one worth fixing and three small ones.

portal-template/src/components/example-layout.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Fumadocs passes a null selector for an operation with `x-exclusiveCodeSample`.
+function ExampleLayout({ selector, usageTabs, responseTabs }: Readonly<LayoutSlots>) {
+  const { schema } = useRenderContext();
+  const { route, examples: bodyExamples, example: defaultId, setExample } = useOperationContext();
+  const pathItem = schema.resolve(schema.dereferenced.paths?.[route]);
+  const operation = pathItem?.[bodyExamples[0].data.method];
+  const examples = useMemo(
+    () => requestExamples(bodyExamples, Parameter.listIn(operation, pathItem, schema.resolve)),
```
saeedjamshaid: **Without SDK samples, the selector does nothing on an operation whose example names come from a parameter.** Here the list comes from the parameter's example names, and `select` passes the choice to fumadocs with `setExample(id)` (line 51). Fumadocs ignores any id that isn't a request-body example, so the cURL tab, which reads fumadocs' selection through `useSelectedRequest`, never changes. Neither does the playground.

Repro: a `GET` whose `view` query parameter has `minimal` and `full` examples, built without `APIMATIC_CODE_SAMPLES_PATH`. The selector offers "Minimal view" and "Full view", but cURL shows `curl -X GET "https://api.example.com/things?view=min"` for both. With SDK samples the same page works, because the SDK tabs read the layout's own selection.

Today every build without the env var hits this. When an operation has no SDK samples, listing only fumadocs' body examples here (what `dev` shows) would avoid it.

portal-template/src/lib/code-samples.server.ts
```diff
@@ -0,0 +1,40 @@
...
+  return file === null ? null : (JSON.parse(await readFile(file, 'utf8')) as SamplesByEndpoint);
+}
+
+/**
+ * Gives every operation the samples of its endpoint, replacing any it already carries. Placed
+ * after bundling, so an operation behind a path-item `$ref` is reached like an inline one.
+ */
+export function placeCodeSamples(document: Document, samples: SamplesByEndpoint | null): Document {
+  const paths = (document as JsonObject).paths;
+  if (samples === null || !isJsonObject(paths)) return document;
```
saeedjamshaid: Minor: this returns early when there is no samples file, so an operation keeps its own `x-apimatic-codeSamples`. With a samples file, the same entry is replaced (efe3e0a). That means one spec renders different samples depending on whether `APIMATIC_CODE_SAMPLES_PATH` is set.

Repro: an operation carrying `x-apimatic-codeSamples` with `OLD_HAND_WRITTEN_SNIPPET` and no catalog entry. Without the env var its TypeScript tab shows `OLD_HAND_WRITTEN_SNIPPET`. With any samples file, the tab is replaced by cURL.

If the catalog is meant to be the only source, stripping the key when `samples` is `null` too would make the two cases match.

src/infrastructure/services/portal-artifacts-service.ts
```diff
@@ -0,0 +1,52 @@
...
+    } catch {
+      return err(ServiceError.invalidResponse(`${describe(resolved)} is not valid JSON.`));
+    }
+    if (!isJsonObject(json)) {
+      return err(ServiceError.invalidResponse(`${describe(resolved)} must hold an object keyed by language.`));
+    }
+
+    const languages = Object.values(Language) as string[];
+    const parsed = Object.entries(json)
+      .filter(([language]) => languages.includes(language))
```
saeedjamshaid: Minor: unknown top-level keys are now skipped without any message. That's the right call for something like `version`, but it also hides a typo. `{ "typescipt": { "paths": { ... } } }` builds a portal with no samples, and nothing says why. A warning naming the ignored keys would catch it.

src/infrastructure/services/portal-artifacts-service.ts
```diff
@@ -0,0 +1,52 @@
+import path from 'node:path';
+import { err, ok, Result } from 'neverthrow';
+import { CodeSampleCatalog, CodeSamples } from '../../types/portal/code-samples.js';
+import { Language } from '../../types/sdk/generate.js';
+import { FilePath } from '../../types/file/filePath.js';
+import { isJsonObject } from '../../types/common/json-object.js';
+import { format as f } from '../../prompts/format.js';
```
saeedjamshaid: Nit: an infrastructure service importing `prompts/format` crosses layers. `.ai/instructions.md` has Prompts and Infrastructure as separate layers, and this is the only such import under `src/infrastructure/`. The service could return the problem and the path, and `prompts/portal/code-samples.ts` could build the message.

portal-template/src/components/example-layout.tsx
mrafnadeem-apimatic: Fixed in 005e95b: without SDK samples the layout lists fumadocs' body examples only, so the selector, cURL and the playground agree. With SDK samples it still lists a parameter's ids; cURL now shows beside them (31a30df) and stays on the body example there, a known inconsistency.

portal-template/src/lib/code-samples.server.ts
mrafnadeem-apimatic: Fixed in 80bb4c7: with no samples file `readCodeSamples` returns an empty catalog, so an operation's own `x-apimatic-codeSamples` are dropped in both cases.

src/infrastructure/services/portal-artifacts-service.ts
mrafnadeem-apimatic: Fixed in 413a1a5: the service returns the keys it skipped, and the CLI warns naming them.

src/infrastructure/services/portal-artifacts-service.ts
mrafnadeem-apimatic: Fixed in 413a1a5: the service returns the file and a `kind`-tagged problem, and `prompts/portal/code-samples.ts` words it.

saeedjamshaid (approved):

PR #355, opened 2026-09-23
feat(portal): give the portal block its v2 shape, with branding, tabs and API options

saeedjamshaid:
> **Stacked on #354.** The base is `saeedjamshaid/split-spec-refs`, so the diff here shows this PR's commits alone. Retarget to `dev` once #354 merges; if it is squash-merged, rebase with `git rebase --onto origin/dev 51e6ccc`.

The `portal` block of `src/apimatic.json` takes the shape it keeps for 2.0: `site`, `brand`, `navigation`, `home`, `api`, `ai` and `advanced`. The portal can be branded with a logo per colour mode, a favicon, a colour preset and primary, fonts, a forced colour mode, a layout and header links. The top level of the site renders as tabs taken from `nav.json`. The spec is `.ai/plans/portal-config.md`, rewritten in the first commit.

## What changes for a user

- **The block is nested.** The flat keys it had on `dev` (`title`, `siteUrl`, `aiPageActions`, …) are now reported as unknown settings, with no hint. That block never shipped. Every problem is reported at once, each by its dotted path.
- **A portal needs a `languages` entry.** At least one is required, for example `"languages": { "typescript": {} }`. Nothing writes it for a portal project yet, so quickstart stops after the scaffold and shows the entry to add rather than starting a preview that would refuse the project.
- **Tabs:**
  - Home, Guides, and the API reference.
  - A top-level folder whose `nav.json` sets `"root": true` gets a tab of its own.
  - `apimatic:pages` is renamed `apimatic:sdks`.
- **The scaffold writes the populated block.** It includes `$schema`, so editors complete and check the file from `apimatic.schema.json`, which ships in the package and covers the whole file.
- **`portal serve` re-applies edits to the `portal` block.** An edit a build would refuse is reported, and the preview keeps what it last accepted. A change to `portal.api`, or a newly created `src/static`, is reported as needing a restart.
- **`portal.api`** sets `groupBy`, `showDeprecated` and `showInternal`. Hidden operations are removed from the document before any page is built, so they appear in no page, sidebar, payload, sitemap or `llms.txt`.

## Commits

One per step of the plan's section 11, then the fixes from a full review of the branch:

1. **The plan**, rewritten for `apimatic.json`.
2. **The nested block:** value objects per namespace, `Color`, the font table, `PortalLanguages`, and name and description taken from the spec.
3. **`apimatic.schema.json`**, with a test that runs the schema and the parser over the same fixtures.
4. **Generated files:**
   - `portal.identity.json` and `theme.css` replace the literal substituted into `portal.ts`.
   - The template applies the site, brand, layout, links, call to action and page actions.
5. **Tabs:**
   - A `root` page-tree transformer builds one root folder per tab.
   - An explicit tab list is passed to the layouts.
   - The CLI validates `root`.
6. **The API options:**
   - The filter runs over the document `bundleSpecification` (#354) returns, before Fumadocs sees it.
   - `groupBy` is passed through.
7. **Serve re-apply:** `FileWatchService` and the serve action.
8. **Scaffold, docs and e2e:** the populated scaffold, README notes, a second e2e build that brands everything, and dated amendments to the plans this one changes.
9. **Review fixes:**
   - **Input validation:** token values, empty static path segments, letter case of files on disk, links that leave the site, CSS colour syntax.
   - **API filter:** `$ref` chains and the keys beside a `$ref`, tags left with no visible operation, route-mode edge cases.
   - **Tabs:** a tab could open on the page with the same path in the tab being left.
   - **Stylesheet:** token overrides now reach the sidebar.
   - **Serve:** the file is read again once the watch starts, saves are coalesced, and a watch that fails says so.
   - **`$schema`:** a prerelease scaffold names its exact version, since jsDelivr's `@2` skips prereleases.

## How it fits #354

`openApiSection` hands Fumadocs `withoutHiddenOperations(await bundleSpecification(file), api)` through the function form of `input`, so there is one server. With #354's inlining, operations behind a path-item `$ref` get pages whatever the filter hides. #354's own test now passes the `api` options.

## Verification

- **`pnpm test`:** 1025 passing, 30 pending. The pending tests are the e2e cases, which need `APIMATIC_E2E=1`. The total includes #354's 18 tests.
- **Lint and type-checks:** ESLint is clean, and both type-checks pass.
- **`APIMATIC_E2E=1`:** 29 passing, 1 skipped as before. There are two real builds:
  - **The default fixture** checks the tabs in the header, the default theme and fonts, and that the browser bundle carries the identity file and nothing from the build-only config.
  - **A branded fixture** uses glass, forced dark, ocean with a primary, Inter with the system mono font, two logos, a favicon, a header link, a call to action on the fallback home page, and a hidden deprecated and internal operation.
- **`theme.css` against Tailwind:** a unit test compiles it for all 11 presets, checking that every override survives and comes after the preset's own rules.
- **Headless Chrome, checked while building:**
  - tabs under all four layouts, with the active tab following the page;
  - forced light and dark modes;
  - the primary colour winning in both modes;
  - a full-width API page under glass;
  - a running `portal serve` picking up a rewritten identity and theme.
- **Sample repository:** the sample in apimatic/sample-docs-as-code-portal resolves and builds with this branch. Its `v2` branch is moved to this layout in apimatic/sample-docs-as-code-portal#12.


src/types/portal/config/link.ts
```diff
@@ -0,0 +1,73 @@
...
+      `'${path}' must be a page of the portal starting with '/', or an address starting with 'https://' or 'http://'.`
+    ]);
+  }
+
+  // Resolved as a browser resolves it, against an origin no link can name: `//host`, `/\host`
+  // and a slash followed by a tab all start with '/' and still lead to another site.
+  private static pageOfThePortal(text: string): string | undefined {
+    try {
+      const resolved = new URL(text, PORTAL_ORIGIN);
+      return resolved.origin === PORTAL_ORIGIN ? `${resolved.pathname}${resolved.search}${resolved.hash}` : undefined;
```
MuHamza30: A header link like /.//evil.com (also /..//evil.com, /%2e//evil.com) passes as an internal page: new URL() keeps the portal origin, but the pathname resolves to //evil.com, so the header renders href="//evil.com", which the browser opens as another host. This is the case the check exists to catch for //host. Could we reject a resolved pathname that starts with //, and add these three inputs to the tests? The link pattern in apimatic.schema.json accepts them too.

portal-template/src/lib/openapi-filter.ts
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * Drops the tags that only removed operations carried, from `tags` and from Redocly's
+ * `x-tagGroups`. Fumadocs builds no folder for such a tag, but every page's payload carries the
+ * document's tags, so a hidden section's name and description would still reach the reader. A
+ * tag another kept tag names as its `parent` stays, as the group it is.
+ */
+function withoutEmptiedTags(root: Json, shown: Json): void {
+  const kept = tagsUsed(shown);
+  const emptied = new Set([...tagsUsed(root)].filter((name) => !kept.has(name)));
```
MuHamza30: emptied only holds tags that operations used, so a group tag with no operations of its own survives when all of its children are hidden. With tags: [{ name: 'Admin', description: 'Internal tools' }, { name: 'Users', parent: 'Admin' }] and every Users operation x-internal, Admin and its description stay in tags, which every page's payload carries. Could a parent whose children were all removed be dropped as well?

src/infrastructure/portal-project-service.ts
```diff
@@ -182,6 +185,47 @@ export class PortalProjectService {
...
+   * its contents change, so an edit that leaves the site as it was, such as a plugin command
+   * rewriting its own block, reloads nothing; the answer says whether anything was written.
+   */
+  public async applyConfig(projectDirectory: DirectoryPath, config: PortalConfig): Promise<Result<boolean, string>> {
+    try {
+      let written = false;
+      for (const [file, contents] of this.appearanceFiles(projectDirectory, config)) {
+        const current = (await this.fileService.fileExists(file)) ? await this.fileService.getContents(file) : null;
+        if (current !== contents) {
+          await this.fileService.writeContents(file, contents);
```
MuHamza30: applyConfig rewrites portal.identity.json and theme.css with writeContents (truncate, then write) while Vite is watching them, so the dev server can pick up an empty or half-written file. FileService.replaceContents() already writes to a temporary file and renames it over the target; could this use it?

src/types/portal/config/fields.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export type Parsed<T> = Result<T, string[]>;
+
+// The URL parser also takes `https:example.com`, which a browser on an https site reads as a
+// path of that site.
+export function isWebAddress(text: string): boolean {
+  return /^https?:\/\//i.test(text) && UrlPath.create(text) !== undefined;
+}
+
+export function isJsonObject(value: unknown): value is Record<string, unknown> {
```
MuHamza30: isJsonObject duplicates the private one in apimatic-config/document.ts:50; could one be exported and reused? Relatedly, METHODS in portal-template/src/lib/openapi-filter.ts:7 includes query, but openapi-slim.ts:15 doesn't, so the two lists already disagree. A single shared list would stop them drifting.

src/types/portal-source-context.ts
```diff
@@ -122,73 +131,165 @@ export class PortalSourceContext {
...
+      );
+      if (match === undefined) {
+        return null;
+      }
+      directory = directory.join(match);
+    }
+    const fileNames = await this.fileService.getFileNames(directory);
+    const match = PortalSourceContext.spelling(
+      fileName,
+      fileNames.map((candidate) => candidate.toString())
```
MuHamza30: spelledOnDisk splits file.relativeTo(...) into strings, maps FileNames through toString(), and rebuilds them with new FileName(match). .ai/skills/value-object.md asks us not to unwrap for path arithmetic and re-wrap. Could the lookup go through FileName/DirectoryPath methods instead?

src/types/portal-source-context.ts
```diff
@@ -122,73 +131,165 @@ export class PortalSourceContext {
...
+   * documents the project's SDK languages; a malformed `plugin` block belongs to the plugin
+   * commands and must not fail a build.
+   */
+  private static parseConfig(
+    document: ApimaticConfigDocument,
+    suggested: SuggestedSite | null
+  ): Result<PortalConfig, PortalSourceProblem> {
+    const block = document.portal();
+    const config = PortalConfig.fromBlock(block, suggested);
+    const languages = PortalLanguages.fromBlock(document.languages(), document.findingsFor('languages'));
```
MuHamza30: parseConfig builds a PortalLanguages value and uses only its errors. PortalLanguages.all()/isPublished() and PortalConfig.siteOrigin(), navigationSettings() and aiSettings() have no callers in src. PortalLanguages.isPublished (portal-languages.ts:60) also re-checks shapes that the languages findings already validated, which goes against "validate once, up front". Could this be a plain function that returns the language errors, with the unused methods removed until something needs them?

MuHamza30 (changes requested):

src/types/portal/config/link.ts
saeedjamshaid: Fixed in d56b7a9 and abb4983. A link with an empty path segment, read as a browser reads it (tabs and newlines dropped, a backslash taken for a slash), is now refused by both the parser and the schema. That one rule covers `//host`, `/\host`, and every dot-segment route to `//host`. The tests have your three inputs plus `/docs/..//host`, `/././/host` and a tab variant, in both `portal-config.test.ts` and the parser-vs-schema table. The two rules agree across a fuzz of about 1.95M short paths.

portal-template/src/lib/openapi-filter.ts
saeedjamshaid: Fixed in b831fbe and 00c1782. A group that no remaining operation carries is now dropped once nothing under it stays, up the chain, and from `x-tagGroups` too. Your Admin/Users example is a test, along with a nested group, a cycle of groups, and a group that stays because a tag under it does.

src/infrastructure/portal-project-service.ts
saeedjamshaid: Done in e4c59f4: `applyConfig` now uses `replaceContents()`. The build-time write stays as it was, since it runs before Vite starts. I checked a real `portal serve` on Windows afterwards, and it still picks up each edit to `apimatic.json`.

src/types/portal/config/fields.ts
saeedjamshaid: Both done. There is now one `isJsonObject` in `src/utils/json-utils.ts`, used everywhere in src that had its own (06a82b4, 6d0a68f), and one in the template. The operation methods are one list in `portal-template/src/lib/openapi-methods.ts`. The drift was a real leak: slim kept any key it did not know as a method, so a sibling `query` operation, and everything under `additionalOperations`, rode along with its schemas in every other page's payload at that path. Slim now drops both (d8c681a, 05b0676).

src/types/portal-source-context.ts
saeedjamshaid: Moved into `FileService.spelledOnDisk` (511e5c9). It walks with the `DirectoryPath` and `FileName` it matched rather than rebuilding them from strings, and the name comparisons now sit at the file-system boundary. The context just calls it.

src/types/portal-source-context.ts
saeedjamshaid: Removed in 8a1ea73: `isPublished`, with the shape re-checks it needed, and the unused `PortalConfig` accessors, plus `siteDescription`, which had no callers either. The tests now read those values through `identity()`, which is what the portal receives. `PortalLanguages` keeps `all()`: #360 reads it to decide the generated SDK pages, so turning it into a plain error function here would only be undone there.

MuHamza30 (approved):

